PR #58, opened 2014-07-14
Preserve cursor position when sending a line or paragraph

munshkr:
When you send a line or a paragraph the cursor moves to the beginning, which is OK, most of the time. But sometimes you may want to test something and not lose track of the current line (especially if it's a function definition spanning multiple lines, or a long line).

I added `g:slime_preserve_curpos` to define if it should preserve the cursor position, because I'm not sure if this is something that you'd want as a default. It makes the whole "sending a paragraph" too silent.

But I'm using vim-slime for livecoding with [Tidal](http://yaxu.org/tidal/) and, in that context, this is a really good feature :)

What do you think?

### DIFF
--- a/doc/vim-slime.txt
+++ b/doc/vim-slime.txt
@@ -137,6 +137,10 @@ g:slime_paste_file	Used to transfer data from vim to GNU Screen or tmux.
 			used. Setting this explicitly can work around some
 			occasional portability issues.
 			whimrepl does not require or support this setting.
+						*g:slime_preserve_curpos*
+g:slime_preserve_curpos	Set to non zero value to preserve cursor position when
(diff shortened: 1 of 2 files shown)


hugovincent:
Any progress on this @munshkr? It'd be great to get this finished and merged.

jpalardy:
I've had my own workaround for this … and it's not pretty.

After consideration, I'm willing to merge this.

@munshkr anything to add?

jpalardy:
The code defaults to _not_ preserving cursor position.

I understand that's the old behavior… but why wouldn't you want this ON?

jpalardy:
Done :-)

PR #68, opened 2015-01-08
Removed conditions for Haskell indentation

### LINKED ISSUE #67: Lines after the first are not indented when "let" is added (Haskell)
```hs
item :: Parser Char
item = Parser (\s -> case s of
    "" -> []
    c:cs -> [(c,cs)])
```
When sending the above to `tmux`, the following error occurs:

```hs
Prelude> :{
Prelude| let item :: Parser Char
Prelude|     item = Parser (\s -> case s of
Prelude|     "" -> []
Prelude|     c:cs -> [(c,cs)])
Prelude| :}

<interactive>:5:5:
    parse error (possibly incorrect indentation or mismatched brackets)
```
The problem is that only the first line is indented when `let` is added.

I took a look at the source and it seems that lines beginning with whitespace are [deliberately not indented](https://github.com/jpalardy/vim-slime/blob/master/ftplugin/haskell/slime.vim#L40). May I ask why this is done? Should I be indenting my code in a different way?

dariusf:
Fixes #67

### DIFF
--- a/ftplugin/haskell/slime.vim
+++ b/ftplugin/haskell/slime.vim
@@ -24,22 +24,15 @@ function! Get_indent_string()
     return repeat(" ", 4)
 endfunction
 
-" indent lines except for first one
+" indent lines except for first one.
+" lines are indented equally, so indentation is preserved.
 function! Indent_lines(lines)
     let l:lines = a:lines
     let l:indent = Get_indent_string()
     let l:i = 1
     let l:len = len(l:lines)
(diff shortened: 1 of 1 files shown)


jpalardy:
Looking good, thanks :-)

dariusf:
I think you may have forgotten to merge this. :)

jpalardy:
Wrong button … :-/

dariusf:
Thanks!

PR #70, opened 2015-02-19
Fixes prompting for config even though g:slime_default_config exists

inside:
Although `let g:slime_default_config = {"socket_name": "default", "target_pane": "1"}` is set in my `~/.vimrc` the plugin keeps on prompting me for a config. This patch fixes this behavior.

### DIFF
--- a/plugin/slime.vim
+++ b/plugin/slime.vim
@@ -127,8 +127,9 @@ function! s:SlimeGetConfig()
   if !exists("b:slime_config")
     if exists("g:slime_default_config")
       let b:slime_config = g:slime_default_config
+    else
+      call s:SlimeDispatch('Config')
     end
-    call s:SlimeDispatch('Config')
   end
 endfunction
 


jpalardy:
Hi there,

The code used to be _exactly_ like that. Then, it was changed to the current mode (ask anyway, answers pre-filled).

There had been some agreement that pressing `enter` twice wasn't too bad.

Now I think that we probably need both, with a new flag (maybe `g:slime_dont_ask_default`).

Thoughts?

inside:
Hi,

I'm not in favor on creating another variable, I think it's not KISS. The readme file should be clearer. Instead of:

> If you do not want vim-slime to prompt for every buffer, you can set a default configuration

I would write:

> If you want vim-slime to prefill the prompt answers, you can set a default configuration

What about it?

jpalardy:
Excellent advice: changed the README's text.

inside:
:thumbsup:

inside:
:thumbsup:

PR #72, opened 2015-03-15
SlimeRunCell

EderSantana:
In Matlab we delimit code blocks with `%%` and run them with `Ctrl+Enter` while inside that block. This PR adds the same functionality to vim-slime with the difference that the user can define her own `g:slime_cell_delimiter` (using each languages comment char). After running the cell, the cursor is moved to the next one. You can run a cell with `<c-c><Enter>`. The user has to define `g:slime_cell_delimiter` to use this. 

This was inspired by the work in [julienr/vim-cellmode](https://github.com/julienr/vim-cellmode) . But there they hard coded the cell delimiter to work with Python.

ToDo:
- [ ] Add ⍐ and ⍗ cell navigation keys 
- [ ] Update docs about this
- [ ] If there is no ending `g:slime_cell_delimiter`, SlimeRunCell should run everything till the last line.
- [ ] Automatically define `g:slime_cell_delimiter` as a function of the language of the current buffer.

### DIFF
--- a/ftplugin/python/slime.vim
+++ b/ftplugin/python/slime.vim
@@ -8,3 +8,4 @@ function! _EscapeText_python(text)
   end
 endfunction
 
+let g:slime_cell_delimiter = "##"
--- a/plugin/slime.vim
+++ b/plugin/slime.vim
@@ -208,7 +208,7 @@ endfunction
 function! s:SlimeRunCell() abort
     " Run a cell delimited by g:cell_delimiter
     call s:SlimeGetConfig()
-    execute ":?" . g:slime_cell_delimiter . "?,/" . g.slime_cell_delimiter . "/y a"
+    execute "silent :?" . g:slime_cell_delimiter . "?;/" . g:slime_cell_delimiter . "/y a"
     "silent :?##?;/##/y a
     ']
     execute "normal! j"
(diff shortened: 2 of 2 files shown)


jpalardy:
Hi there,

It's looking good -- I'm not completely opposed to this, but I'm wondering, in practice, when this is a better solution than breaking "cells" into different paragraphs. (paragraph selection is vim-slime's default behavior)

The example from vim-cellmode didn't give me more insights.

EderSantana:
I was mostly thinking about running code that comes from the [IPython/Jupyter](http://nbviewer.ipython.org/) project. They have notebooks for several languages and block codes are separated with cells. If we separate the blocks with the same cell marks, it can be exported to the notebook format as well. 
But as you said, we can simply use paragraphs for that and add the cell marks in the first line of the paragraph. And the only contribution of this would be for running code imported from such projects. 

Another use that I had in mind was allowing for code breakpoints. Having sections of several paragraphs in a script and run one at time. Ex. section of ETL, section of training, section of testing. Then we mark each section with a header and save a few keystrokes...

Anyways, I don't mind if you want to let this hanging for a while to see if there is any further insights first.

jpalardy:
no activity, closing.

PR #73, opened 2015-03-18
Add support for lhaskell, use filter() in Remove_line_comments.

dbeecham:


jpalardy:
Hi there,

The changes are looking good, although I'm a bit worried about the code duplication between this file and the haskell one.

Is there are a way to merge those files into one?

dbeecham:
There we go, files are merged. lhaskell is now a symbolic link to lhaskell, and haskell's file has logic for both.

jpalardy:
Awesome, thanks

PR #74, opened 2015-03-24
In Haskell don't prepend let if line starts with colon

werbitt:
This allows for sending commands to GHCi


jpalardy:
thank you :-)

PR #75, opened 2015-04-03
haskell: newtype is not prefixed

dbeecham:


jpalardy:
Looking good, thanks.

PR #77, opened 2015-09-11
Adjusting behaviour to only use cpaste magic on multi-line code trans…

oddraisin:
…mission.

A small change so that, when sending a single line of text, don't use cpaste. Useful if you want to send a single line (with a potential logic error in it) to ipython, then use up/history in ipython to play with the line. Without this, every line is sent with %cpaste magic, which makes the code unrecoverable within the ipython session.


jpalardy:
Looks good to me.

Thanks :-)

PR #78, opened 2015-09-23
Add support for sml

andars:
Not sure if you want this in the mainline, but I'd thought I'd give you the option.


jpalardy:
Looks good, thanks for that.

PR #79, opened 2015-10-14
force overwrite of paste-file when noclobber is set for shell

Orbots:


jpalardy:
Hi there,

It's unclear that `>!` would work in all shells. What's your setup? Can you refer me to supporting documentation?

Thanks.

Orbots:
Ah yeah.  Looks like this is isn't the case for all shells.  For example, Bash uses >| instead.

https://www.gnu.org/software/bash/manual/bash.html 
section 3.6.2

Bourne shell doesn't have a concept of noclobber.

I'm using tcsh in CentOS 6.

http://unix.stackexchange.com/questions/45201/bash-what-does-do

Seems like a full solution needs quite a few conditionals.  

Alternatively the paste-file could be deleted ( rm -f ) first before the cat >

I'll close this since it's not a correct solution for everyone.

Thanks,

- Michael

jpalardy:
Thanks for the clarifications.

Issue #69, opened 2015-02-11
Vim Slime not using slime_default_config values


ethagnawl:
My .vimrc contains:

`let g:slime_default_config = {"socket_name": "default", "target_pane": "1"}`

Yet when I attempt to evaluate a line of code, I'm still prompted for a socket and target pane. (I should note that the values listed above _are_ prepopulated in the prompt, but I was expecting to not have to manually accept each of them whenever I eval code in a new buffer.)

jpalardy:
Hi there,

Prepopulated is the "correct" behavior. I'm sorry for the confusion. You can review [this](https://github.com/jpalardy/vim-slime/issues/62) too.

ethagnawl:
Got it - thanks for clearing that up. 

If I come up with a better way to word what's in the README, I'll submit a PR.

Anyways, thanks for the prompt response and the great plugin!

jpalardy:
I'd certainly welcome a README PR, thanks :-)

Issue #62, opened 2014-10-08
How to automatically set a target pane?


krzkrzkrz:
Was `let g:slime_default_config = {"socket_name": "default", "target_pane": "1"}
` removed? Cant see anything in the help file

Moreover, it keeps prompting me for the first time, to select socket and target pane. Normally, with above configuration in vimrc, vim-slime wouldn't need to prompt.

PR #80, opened 2015-10-17
adding slime_dont_ask_default config option which bypasses config prompt

ethagnawl:
Per issues #62, #69 (opened by me) and pull request #70, it seems like this is an option people want. (My approach was directly inspired by @jpalardy's comments in PR #70.)

It might make sense to validate the presence of the required options for the various multiplexers inside their respective *Config functions, but I thought this was enough to get the conversation started.

### DIFF
--- a/README.md
+++ b/README.md
@@ -86,6 +86,10 @@ If you want vim-slime to prefill the prompt answers, you can set a default confi
 
     let g:slime_default_config = {"socket_name": "default", "target_pane": "1"}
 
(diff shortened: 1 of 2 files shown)


jpalardy:
Hi there,

Thanks for the PR. I integrated your fix but I tweaked the details:

- I made the slime_dont_ask_default flag a global, for consistency (updated the README as such)
- I flattened the logic in that function, it was getting hard to understand

What do you think?

ethagnawl:
That all looks fine by me. Thanks!

PR #82, opened 2015-10-20
Matlab eval wrap

maksay:
When sending a code to matlab, it is often hard to distinguish between the code sent and the output produced. Furthermore, if sending a long list of instuctions, errors produced by the first ones quickly get burried under the rest of the code pasted. Additionally, there is no visual distinction between the errors and the output in terminal as compared to the matlab gui. Finally, when part of the code produced errors, you may want to stop the execution and not to produce any additional output.

Here is a simple solution for aforementioned problems that sends the code wrapped around in eval.  Behaviour can be toggled by setting / unsetting g:slime_matlab_eval. It supports escaping single quotes, which are used most often.

### DIFF
--- a/doc/vim-slime.txt
+++ b/doc/vim-slime.txt
@@ -142,7 +142,7 @@ g:slime_paste_file	Used to transfer data from vim to GNU Screen or tmux.
 g:slime_python_ipython	Set to non zero value when using IPython to enable
 			%cpaste support in Python buffers.
 
-                                                g:slime_matlab_eval
+                                                *g:slime_matlab_eval*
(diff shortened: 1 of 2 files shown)


jpalardy:
PR was closed?

What happened?

maksay:
Closed as I found a really weird behaviour when you have somewhere in the text a newline, then single quote, followed by two tabs - matlab prints out all available function names. Hopefully fixed it now, will test some more and if it works, will issue a new request, if that is OK.

jpalardy:
Sounds good, thanks.

PR #83, opened 2015-10-20
Matlab eval wrap

maksay:
When sending a code to matlab, it is often hard to distinguish between the code sent and the output produced. Furthermore, if sending a long list of instuctions, errors produced by the first ones quickly get burried under the rest of the code pasted. Additionally, there is no visual distinction between the errors and the output in terminal as compared to the matlab gui. Finally, when part of the code produced errors, you may want to stop the execution and not to produce any additional output.

Here is a simple solution for aforementioned problems that sends the code wrapped around in eval and sprintf. Behaviour can be toggled by setting / unsetting g:slime_matlab_eval. It supports escaping single quotes and printf-style formatted strings.

### DIFF
--- a/doc/vim-slime.txt
+++ b/doc/vim-slime.txt
@@ -142,7 +142,7 @@ g:slime_paste_file	Used to transfer data from vim to GNU Screen or tmux.
 g:slime_python_ipython	Set to non zero value when using IPython to enable
 			%cpaste support in Python buffers.
 
-                                                g:slime_matlab_eval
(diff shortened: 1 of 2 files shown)


jpalardy:
This is quite an involved piece of code.

Could you take a screenshot of the problem, and maybe another one of how you fixed it?

I would also like an explanation of the code.

Thanks :-)

jpalardy:
no responses

Issue #66, opened 2014-12-29
Can't SlimeSend1 from Haskell file without "let " prefix


gfixler:
I know there's a Haskell ftplugin file that tries to add `let`s to the appropriate lines, and it does pretty well, but if I put `main` on its own line and send it, or if I try to `:SlimeSend1 main` to run the main function, e.g., I can't get it to send without the prefix. Is there a way to send something raw from the command line, and/or the buffer? I also couldn't set up a mapping to let me send `runhaskell %` to call runhaskell on the whole file to a shell pane, because of the let. As a simple workaround while sending to ghci, I can prefix what I send with `() = () in`, which eats the let, but it's less than ideal.

ethagnawl:
@jpalardy @gfixler I took a pass at this in PR #84.

PR #84, opened 2015-11-09
adding `main` to not_prefixable_keywords in ftplugin/haskell/slime.vim

ethagnawl:
Addresses the following portion of #66: 

> I know there's a Haskell ftplugin file that tries to add lets to the appropriate lines, and it does pretty well, but if I put main on its own line and send it, or if I try to :SlimeSend1 main to run the main function, e.g., I can't get it to send without the prefix.

2015-11-09-004337_1600x900_scrot


jpalardy:
looks good to me :-)

thanks.

PR #89, opened 2015-12-05
Fixed issue where python code was not being executed when sent to python interpreter.

jinyeow:
I was having issues with code sent to the python interpreter getting stuck after for loops or if's. It needed an extra newline for the interpreter to see that the code block had finished and to go ahead and execute. 

The change is on line 7 where I simply append a newline. The extra "\n" fixes the issue so that the code sent to the python interpreter is executed.


jpalardy:
Damn… ok, thanks for this :-)

Issue #85, opened 2015-11-15
Unwanted let prefixing when using vim slime with Haskell


lpil:
Hello!

I'm trying to use vim slime with Haskell and the Tidal library to programatically make drum beats live. I wish to evaluate expressions like these:

```haskell
d1 <- dirtStream

d1 $ sound "bd"
```

But when I try to do with with vim-slime, I get this:

```haskell
let d1 <- dirtStream

let d1 $ sound "bd"
```

Neither of these expressions are valid.

How do I work around this?

Thanks,
Louis

jpalardy:
I'll refer to the Haskell plugin authors:

@ethagnawl @romeovs @dbeecham @werbitt @dariusf @jdiez17 thoughts?

lpil:
Thanks Jonathan

dbeecham:
What is the rationale for using not_prefixable_keywords? I'm trying to wrap my mind around when to prepend 'let'.

Do we need to search for an equal sign (=) before prepending?

This needs to be prepended
    f x = 3*x

but not this
    f <- x

and not this
    x

ethagnawl:
@lpil Are you able to wrap those expressions in a `do`? That would solve the `let` problem.

lpil:
No, I am not I'm afraid.

lpil:
This is still quite the show stopper for me here. If we cannot fix this feature in the near future, is there at least a way to disable it?

dbeecham:
Yes, in ftplugin/haskell/slime.vim, in _EscapeText_haskell, comment out the
line with 'Perhaps_prepend_let'.

lör 21 nov 2015 15:20 Louis Pilfold <notifications@github.com> skrev:

lpil:
Thank you.

PR #95, opened 2016-03-14
Haskell: Avoid incorrect let prefixing

lpil:
I've altered the logic that determines whether a statement should be prefixed with `let ` so that it only applies to statements that are likely to be assignments, as is desirable.

Previously this feature resulted in incorrect statements being sent to GHCi.

i.e. `1 + 1` became `let 1 + 1`

Fixes https://github.com/jpalardy/vim-slime/issues/85

### DIFF
--- a/ftplugin/haskell/slime.vim
+++ b/ftplugin/haskell/slime.vim
@@ -1,5 +1,3 @@
-let s:not_prefixable_keywords = [ "import", "data", "instance", "class", "{-#", "type", "case", "do", "let", "default", "newtype", "foreign", "--", "main"]
-
 " Remove '>' on line beginning in literate haskell
 function! Remove_initial_gt(lines)
     return map(copy(a:lines), "substitute(v:val, '^>[ \t]*', '', 'g')")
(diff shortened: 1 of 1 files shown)


jpalardy:
At this point, I don't have enough information to know what we'd lose if we merged this change…

maybe @ethagnawl @romeovs @dbeecham @werbitt @dariusf (from git blame) might want to chime in?

dbeecham:
This does not work if you slime

    f :: Int -> Int
    f x = 3 + x

since it only checks the first line for the equal sign. Maybe check for "::" on line 1 as well; if there is an "::" on line 1 we're basically guaranteed to have a function definition following (I think?)

lpil:
I think that is sensible.

lpil:
Hey, I've updated the code to handle type signature definitions as suggested by @dbeecham :)

dbeecham:
This merge request has my thumbs up at this point. :)

jpalardy:
Haskellers know best :-D

Let me know if it causes problems.

lpil:
Woohoo! Thanks!

PR #100, opened 2016-05-26
Use absolute path to tmux socket

### LINKED ISSUE #99: vim-slime should ask for a socket path, not a socket name
tmux supports two ways to pick the session's socket from the command line:

`-L` : User passes a name of a socket that exists in `$TMPDIR`
`-S` : User passes an absolute path to a socket

On shared hosts, it is not safe to have tmux putting its sockets into the default value for `$TMPDIR`, `/tmp`, because they are likely to be blown away. However, it is often undesirable to set `$TMPDIR` to a path in the home directory because other processes can fill it up with garbage. tmux switched to a separate variable, `$TMUX_TMPDIR`, but not until recently, some time after the latest version in use at my company (1.8). Altogether, it makes the use of `-L` very hard to get right. For example, vim-slime simply cannot function when I use 1.8, keep my tmux socket in my home directory on a shared machine, but leave `$TMPDIR` unset or pointing at `/tmp`.

To get around this issue, I think it would be better if vim-slime asked for a full path to the socket and used `-S` internally. This will require plugin users to type more than a socket name, but I imagine the vast majority of users are sending snippets to the same session, not a different session. The current session's socket path is kept in `$TMUX` and can be parsed in vim with `split($TMUX, ',')[0]`. It can be used as the default path and fit the needs of most users [^1]. I have made [this improvement before for vim-tmux-navigator](https://github.com/christoomey/vim-tmux-navigator/pull/88/commits/5d76881c6ee5de89994ef7e2308bb1ddaef38bea). I will be happy to volunteer a patch this evening, but it will be your decision whether to accept it.

[^1]: I honestly suspect it will fit the need of every user, and thus the prompt can be removed entirely, but I might be wrong.

thejohnfreeman:
Pull default from `$TMUX` in environment.

Closes #99

I also changed the default pane to `:.2` instead of the current pane. If that isn't acceptable, then I think it makes sense to split the config into independent variables instead of packaging them in a struct, so that their defaults can be overridden individually.

### DIFF
--- a/plugin/slime.vim
+++ b/plugin/slime.vim
@@ -48,26 +48,34 @@ endfunction
 " Tmux
 """""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""
 
+function! s:TmuxSocket()
+  " The socket path is the first value in the comma-separated list of $TMUX.
+  return split($TMUX, ',')[0]
+endfunction
+
+function! s:TmuxCommand(args)
+  return 'tmux -S ' . s:TmuxSocket() . ' ' . a:args
+endfunction
+
 function! s:TmuxSend(config, text)
-  let l:prefix = "tmux -L " . shellescape(a:config["socket_name"])
   " use STDIN unless configured to use a file
   if !exists("g:slime_paste_file")
-    call system(l:prefix . " load-buffer -", a:text)
+    call system(s:TmuxCommand("load-buffer -"), a:text)
   else
     call s:WritePasteFile(a:text)
-    call system(l:prefix . " load-buffer " . g:slime_paste_file)
+    call system(s:TmuxCommand("load-buffer " . g:slime_paste_file))
   end
-  call system(l:prefix . " paste-buffer -d -t " . shellescape(a:config["target_pane"]))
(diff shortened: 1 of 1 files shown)


plugin/slime.vim
```diff
@@ -44,26 +44,34 @@ endfunction
 " Tmux
 """""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""
 
+function! s:TmuxSocket()
+  " The socket path is the first value in the comma-separated list of $TMUX.
+  return split($TMUX, ',')[0]
```
jpalardy: The presence of the TMUX environment variable assumes that we are running VIM inside TMUX. That's not necessarily the case…

thejohnfreeman: If they're configuring slime for use with tmux, I would presume it is, right?

jpalardy: The destination terminal needs to run TMUX. The "source" terminal, where VIM runs, doesn't. In practice, I've used vim-slime in both configurations.

However the configuration can be obtained, I don't think it's a good idea to assume the (TMUX) variable exists.

jpalardy:
Closing this -- not useful for everybody.

PR #101, opened 2016-06-02
Add support for Conemu

selliera:
[ConEmu](http://conemu.github.io/) is a windows console emulator with many features, which is also the base for the [cmder](http://cmder.net/) console emulator for windows.
Support for ConEmu enables one to benefit from vim-slime under windows, without requiring cygwin or msys.

Use conemuc to send text to the selected conemu console (default to the
active tab of the first found conemu window).
Use the windows clipboard to send text, so store and restore the
previous value.

### DIFF
--- a/plugin/slime.vim
+++ b/plugin/slime.vim
@@ -79,7 +79,9 @@ endfunction
 
 function! s:ConemuSend(config, text)
   let l:prefix = "conemuc -guimacro:" . a:config["HWND"]
-  " use STDIN unless configured to use a file
+  " use the selection register to send text to ConEmu using the windows
+  " clipboard (see help gui-clipboard)
+  " save the current selection to restore it after send
   let tmp = @*
   let @* = a:text
   call system(l:prefix . " print")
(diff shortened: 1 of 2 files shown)


plugin/slime.vim
```diff
@@ -74,6 +74,27 @@ function! s:TmuxConfig() abort
...
+" Conemu
+"""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""
+
+function! s:ConemuSend(config, text)
+  let l:prefix = "conemuc -guimacro:" . a:config["HWND"]
+  " use STDIN unless configured to use a file
+  let tmp = @*
+  let @* = a:text
+  call system(l:prefix . " print")
+  let @* = tmp
```
jpalardy: `@*` is a trick I don't know. Can you explain this code?

jpalardy:
The code is simple enough, but there's a bunch of places where documentation would need to be taken into consideration.

selliera:
You are right, the comments in my patch are just plain wrong; I will improve that.

@* is the way to interact with the * (selection) register in vim. Under windows, selection (`*`) and drop (`+`) registers are identical, and use the system clipboard.

So `let tmp = @*` saves the current state of the clipboard, `let @* = a:text` fills the windows clipboard with the text to send. The subsequent `system` command uses a conemu command to send the content of the clipboard to a given console. Then `let @* = tmp` restores the content of the clipboard, so users C-c/C-v are not affected.

selliera:
The readme has been updated to mention conemu, and there are more comments in the code, to ease maintainability.

jpalardy:
Thank you :-)

PR #103, opened 2016-06-04
use getpos('.') instead of getcurpos() for old vim

selliera:
the getcurpos function doesn't exist before 7.4.313, so work around the
issue if it is not available.


jpalardy:
Oh, good catch. Thanks.

PR #106, opened 2016-06-25
python escape text: dedent a code block starting at non-zero level

katusk:
- More robust empty line removal (especially at the beginning of text)
- Only adding new lines at the end of text when needed
- Dedent a code block starting at non-zero level

### DIFF
--- a/ftplugin/python/slime.vim
+++ b/ftplugin/python/slime.vim
@@ -3,7 +3,8 @@ function! _EscapeText_python(text)
   if exists('g:slime_python_ipython') && len(split(a:text,"\n")) > 1
     return "%cpaste\n".a:text."--\n"
   else
-    let no_empty_lines = substitute(a:text, '\(^\|\n\)\zs\s*\n\+\ze', "", "g")
+    let empty_lines_pat = '\(^\|\n\)\zs\(\s*\n\+\)\+'
+    let no_empty_lines = substitute(a:text, empty_lines_pat, "", "g")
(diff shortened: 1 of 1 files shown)


jpalardy:
Hi there,

That code is getting quite complicated :-)

I don't mind merging those as long as nobody complains. Moving forward, will you be available to support it, if needed?

katusk:
Yeah, just a bit complicated, some horrible escaping is needed both with the vim regexes and with the python interpreter :) I do not plan to add anything else, I think these were the finishing touches from me.

You basically need to do three things, when sending to the python interpreter, and you do not really need anything else --- these are all fully ticked now, I think:

1. Squeeze. Get rid of all lines with whitespace only as a prefiltering. This greatly simplifies the next steps.
2. Dedent the whole text as much as possible in order to send code blocks beginning at a non-zero indentation level.
3. Close blocks. Add new lines where needed to close code blocks at the zero indentation level.

Sure, I am around, just add a mention, so I will be directly notified. And if anybody can simplify those regexes, I would be most happy.

jpalardy:
Sounds good, thanks :-)

katusk:
You are welcome. I have an additional custom-made .py file (originally Unix EOL, UTF-8 w/o BOM) littered with crazy but still somewhat reasonable whitespaces as a test file, I attach it here: [badlines.py.txt](https://github.com/jpalardy/vim-slime/files/333424/badlines.py.txt)

PR #107, opened 2016-06-30
Do not prepend "let" when sending type definitions

clarkenciel:
Currently various type declarations will fail since `let` is prepended. This PR adds a function to check whether a line is a `newtype` declaration, a type alias, or a datatype declaration. That function is then added to `Perhaps_prepend_let`'s check so that `let` is not prepended to such declarations.

NB: This works when one presses C-c, C-c on an individual type declaration. There is more work to be done for submitting blocks or entire files


jpalardy:
looks good -- thanks :-)

Issue #90, opened 2015-12-09
Feature for putting output back into vim


xdl:
Loving using this plugin, although I was wondering if there's an easy way to send the results of the evaluation back into the vim buffer automatically? I'm using this with tmux at the minute and am finding that switching over to the tmux pane, copy the output, switching back to paste it in getting a bit tedious.

jpalardy:
Hi there,

"easy way" … I don't think so, unless I've missed something. When we paste text to tmux, it just flushes it in the input buffer (as if you had typed it yourself)… but what happens (the feedback) is a result of whatever is running.

I can imagine doing a snapshot before, a snapshot after, and diff both -- taking that as the "result" of the command you just ran. I'm not sure how easy/hard that would be … or whether it would be useful at all.

In what context do you need to copy the result back in vim? (maybe there are other ways)

xdl:
Ah I see, yeah there weren't any easy ways that came to my mind either.

I have a ipython session pane open alongside a vim pane in tmux, and I've gotten into the habit of pasting the results of snippets of python evaluations back into vim to remind me of how I developed a function (kind of like really lazy tdd).

Maybe wrapping the contents of what you send into tmux with a session (bash, ipython, ghci etc.) specific callback that sends the output back to tmux to give back to the vim pane? Does that sound like something that might work - I don't mind spending some time looking into this if you think it's plausible!

Eddie

jpalardy:
You can feed and read back a command with the `system(…)` function. If you do that, you need to feed enough context to run… it's a new process every time. I mean: feed the function and feed the invocation, both.

It's similar to what happens when you select a block of text and feed it to python:

    print 1 + 2   # shift-v select, !python,  => 3

xdl:
Hmm do you think there might be a way to preserve the environment of the interpreter (e.g. previously defined variables etc.) between `system` calls?

jpalardy:
no, not really :-(

PR #108, opened 2016-07-03
Tmux eval result in buffer

xdl:
Hi, I had go addressing this [issue](https://github.com/jpalardy/vim-slime/issues/90) I brought up a while back, regarding putting the results of the evaluation back into vim with a keybinding. I've had a play around with the feature and it's working nicely on tmux 1.8 and 2.2, using the snapshot idea you mentioned. The only kludge is erasing the history of the pane (get around the edge condition of the max history length being reached) when `<C-c><C-c>` is pressed. Let me know what you think!

Eddie

### DIFF
--- a/README.md
+++ b/README.md
@@ -130,6 +130,21 @@ tmux target pane:
             (either session name or number), the ith window and the jth pane
     "%i"    means i refers the pane's unique id
 
+#### Get Difference
+
+This is an experimental feature for putting the evaluation results back into the current vim buffer, analogous to supplying a prefix argument for evaluation in SLIME or Geiser. Enable with:
+
+    let g:slime_take_snapshot = 1
+
+`<C-c><C-d>` will put the difference between what was in the target buffer before the last time text was sent (`<C-c><C-c>`), and the current state.
+
+The following parameters are configurable:
+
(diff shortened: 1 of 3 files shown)


jpalardy:
Hi there,

If this weren't vimscript, I would re-organize the code. For this specific case, I would like "events" like beforeSend, or afterSend (etc).

I wouldn't mind keeping this code on a branch, but I'm not sure I want to add it to master.

What do you think?

xdl:
Yep, agree that the feature switch checking should be moved to a beforeSend event handler. I'm happy to leave it on a branch - I appreciate that it's niche feature, although I find myself having to do the equivalent tmux copy motions quite regularly.

Actually, I realised that there needs to be a bit more logic involved in the GetDifference so hold off on merging it yet. Do you see any issue in moving the GetDifference function into a shellscript file for readability?

jpalardy:
Since it's going on a branch, feel free to implement `GetDifference` however you like :-)

I'll be gone for a few days. I'll take a look when I'm back.

jpalardy:
There, I created a branch `getDifference` with your commits

jpalardy:
you're merged on the getDifference branch

jpalardy:
This lives on the branch, closing.

Issue #109, opened 2016-07-11
Ipython 5.0.0 in tmux  - Have to send lines twice


nigeljyng:
I'm using Vim-slim with IPython 5 in tmux. Before this, I can send blocks of text by highlighting text in visual mode, then doing the Ctrl-C, Ctrl-C command. But since the upgrade to IPython 5, Ctrl-C, Ctrl-C only seems to enter pasting mode, without actually pasting the text. I have to do the Ctrl-C sequence twice for the text to paste.

Here's a few screenshots to show what I mean:

### Start:
![1](https://cloud.githubusercontent.com/assets/13656877/16725335/ab0ba026-474e-11e6-9660-e88cce418855.png)

### If I send a single line, things work as usual:
![screen shot 2016-07-11 at 10 04 06](https://cloud.githubusercontent.com/assets/13656877/16725371/dc8baa42-474e-11e6-85cd-df4dfa4d54fc.png)

### But if I highlight a few lines and hit send, it seems to just enter %cpaste mode, without sending anything:
![screen shot 2016-07-11 at 10 04 52](https://cloud.githubusercontent.com/assets/13656877/16725399/f61aa698-474e-11e6-854d-04de207520b3.png)

### I have to hit Ctrl-C, Ctrl-C again to send lines:
![screen shot 2016-07-11 at 10 05 55](https://cloud.githubusercontent.com/assets/13656877/16725423/17560b68-474f-11e6-9fd1-4e6de1574a53.png)

Any chance this can get a fix?

jpalardy:
Hi there,

Just to be clear: are you saying everything worked in iPython < 5 and now it doesn't?

nigeljyng:
Yes. Everything worked on IPython 4.x. I only encountered this problem after upgrading to IPython 5.

vrivellino:
@jpalardy (cc @nigeljyng): I submitted a small PR to fix this issue. It looks like sending `%cpaste` separately from the code to IPython fixes this issue.

jpalardy:
merged: thanks to @vrivellino .

Let me know if it's still a problem.

rsmith31415:
@jpalardy Are you sure this fixes the issue? I updated with Vundle and verified that the latest PR was included, but I still see the same behavior. 

Using neovim 0.1.5-dev, tmux 2.1, Ubuntu 16.04

jpalardy:
@vrivellino thoughts?

vrivellino:
[...]

PR #110, opened 2016-07-15
Break python escape into pieces for IPython

vrivellino:
Fix for #109

Sending `%cpaste` separately from the code seems to fix this for me. I did a quick test against IPython 4.2.1 (as well as 5) to make sure this didn't break things for IPython 4 users.


jpalardy:
looking good, thanks :-)

PR #115, opened 2016-10-19
change python escape to fix IPython 5.1 output

ArcturusB:
When sending text to IPython 5.1, the colons that should be leading every line of code are all printed at the end of the input. The Python output (`Out[]:` or the result of a `print`) is printed after these columns without any newline. 

slime_ipython_before

Not only is this output ugly, but more importantly it can become difficult to read when dealing with a large number of lines. 

The current pull request addresses this issue, resulting in a cleaner output.

slime_ipython_after


jpalardy:
works for me, thanks :-)

PR #116, opened 2016-11-11
send current line

kindlychung:
Added a new function and a corresponding command to send the current line to repl.

### DIFF
--- a/plugin/slime.vim
+++ b/plugin/slime.vim
@@ -257,6 +257,11 @@ function! s:SlimeSend(text)
   endfor
 endfunction
 
+
+function! s:SlimeSendCurrentLine()
+  call s:SlimeSend(getline(".") . "\r")
+endfunction
+
 function! s:SlimeConfig() abort
   call inputsave()
   call s:SlimeDispatch('Config')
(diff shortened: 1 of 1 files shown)


plugin/slime.vim
```diff
@@ -277,6 +282,7 @@ command -bar -nargs=0 SlimeConfig call s:SlimeConfig()
 command -range -bar -nargs=0 SlimeSend <line1>,<line2>call s:SlimeSendRange()
 command -nargs=+ SlimeSend1 call s:SlimeSend(<q-args> . "\r")
 command -nargs=+ SlimeSend0 call s:SlimeSend(<args>)
+command! SlimeSendCL call s:SlimeSendCurrentLine()
```
jpalardy: How about inlining the whole line here?

    command! SlimeSendCurrentLine call s:SlimeSend(getline(".") . "\r")

jpalardy:
How do you use this in practice?
Do you have a key binding to invoke this?
Is it better than `shift-v ctrl-c, ctrl-c`?

plugin/slime.vim
kindlychung: That's also fine with me.

kindlychung:
I mapped it to <c-c><c-l>, and yes, the experience is significantly better.

plugin/slime.vim
jpalardy: If you could update the PR, then I'll merge it.

jpalardy:
great, thanks :-)

PR #117, opened 2016-11-14
Correct Neovim

jrpotter:
Neovim does not respect usage of g:slime_paste_file. Correcting.


jpalardy:
looking good, thanks :-)

PR #121, opened 2016-12-26
Replace multiple system calls in function ScreenSend

sindex:
By using the eval command, Screen is run once per function call.


jpalardy:
Good stuff, thanks.

PR #124, opened 2017-01-20
Update slime.vim

raubreywhite:
When trying to run your version of slime.vim, I had one window to the left (bash), and the vim window to the right. On trying to send the simple command "ls" (or any command) from the right window to the left, it would always end up sending it to the vim window, not the bash window.

By separating out the "sendscreen" command into the four separate calls, I managed to fix this error.

I found this error to be present on:
1. macOS Sierra 10.12 with standard Mac VIM + https://github.com/FreedomBen/screen-for-OSX
2. In this Dockerfile: https://github.com/raubreywhite/docker/tree/master/dpython

(Note in the linked dockerfile I install my forked version of vim-slime. If I install your version, then the error is present).

### DIFF
--- a/plugin/slime.vim
+++ b/plugin/slime.vim
@@ -27,7 +27,13 @@ end
 function! s:ScreenSend(config, text)
   call s:WritePasteFile(a:text)
   call system("screen -S " . shellescape(a:config["sessionname"]) . " -p " . shellescape(a:config["windowname"]) .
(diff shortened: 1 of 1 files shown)


jpalardy:
Collapsing multiple calls to one line was a recent change. Let's split it back again.

Thanks for the PR.

PR #127, opened 2017-02-09
Work with absolute path to tmux socket

### LINKED ISSUE #99: vim-slime should ask for a socket path, not a socket name

thejohnfreeman:
- Pull default socket path from `$TMUX` environment variable, if it exists.
- Pass -S or -L to tmux command depending on whether the socket path is
  absolute or relative.

Closes #99

### DIFF
--- a/plugin/slime.vim
+++ b/plugin/slime.vim
@@ -20,6 +20,14 @@ if !exists("g:slime_paste_file")
   let g:slime_paste_file = expand("$HOME/.slime_paste")
 end
 
+if !exists("g:slime_default_tmux_socket_name")
+  let g:slime_default_tmux_socket_name = "default"
+end
+
+if !exists("g:slime_default_tmux_target_pane")
+  let g:slime_default_tmux_target_pane = ":"
+end
+
 """""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""
 " Screen
 """""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""
@@ -52,22 +60,13 @@ endfunction
 " Tmux
 """""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""
 
-function! s:TmuxDefaultSocket()
-  " The socket path is the first value in the comma-separated list of $TMUX.
-  return empty($TMUX) ? "default" : split($TMUX, ",")[0]
-endfunction
-
-function! s:IsAbsolute(path)
-  " Case sensitivity does not matter here, but let's follow good practice.
-  " TODO: Make this cross-platform. Windows supports tmux as of mid-2016.
-  return a:path[0] ==? "/"
-endfunction
-
 function! s:TmuxCommand(config, args)
   let l:socket = a:config["socket_name"]
-  " Use tmux -S for an absolute path to the socket.
-  " Use tmux -L for a relative path to the socket in tmux's temporary directory.
-  let l:socket_option = s:IsAbsolute(l:socket) ? "-S" : "-L"
+  " For an absolute path to the socket, use tmux -S.
+  " For a relative path to the socket in tmux's temporary directory, use tmux -L.
+  " Case sensitivity does not matter here, but let's follow good practice.
+  " TODO: Make this cross-platform. Windows supports tmux as of mid-2016.
+  let l:socket_option = l:socket[0] ==? "/" ? "-S" : "-L"
   return system("tmux " . l:socket_option . " " . shellescape(l:socket) . " " . a:args)
 endfunction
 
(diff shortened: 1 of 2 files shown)


thejohnfreeman:
My second attempt at this solution, which I believe addresses your previous concerns.

plugin/slime.vim
```diff
@@ -52,22 +52,41 @@ endfunction
...
 function! s:TmuxPaneNames(A,L,P)
   let format = '#{pane_id} #{session_name}:#{window_index}.#{pane_index} #{window_name}#{?window_active, (active),}'
-  return system("tmux -L " . shellescape(b:slime_config['socket_name']) . " list-panes -a -F " . shellescape(format))
+  return s:TmuxCommand(b:slime_config, "list-panes -a -F " . shellescape(format))
 endfunction
 
 function! s:TmuxConfig() abort
   if !exists("b:slime_config")
-    let b:slime_config = {"socket_name": "default", "target_pane": ":"}
+    let b:slime_config = {"socket_name": s:TmuxDefaultSocket(), "target_pane": ":.2"}
```
jpalardy: `.2` ?

jpalardy:
I don't see myself merging this.

You can set the slime config in your .vimrc at startup.

Alternatively, I can imagine the whole prompting-logic to be a function -- users could pass a function pointer for how they want the prompting to be done.

thejohnfreeman:
Ay, caramba. It's a useful addition to set more reasonable defaults in more situations that disturbs nothing else. What does it take to get patches into this project? If you want me to remove the default pane, that's fine, but it's hard to do that when you just close the PR.

plugin/slime.vim
```diff
@@ -52,22 +52,41 @@ endfunction
 " Tmux
 """""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""
 
+function! s:TmuxDefaultSocket()
```
jpalardy: I don't think we need this function, this logic could be done by the user.

plugin/slime.vim
```diff
@@ -52,22 +52,41 @@ endfunction
...
+  " The socket path is the first value in the comma-separated list of $TMUX.
+  return empty($TMUX) ? "default" : split($TMUX, ",")[0]
+endfunction
+
+function! s:IsAbsolute(path)
```
jpalardy: can you inline this function? around line 69?

plugin/slime.vim
jpalardy: this follows my comment from line 55

jpalardy:
I had another look, I think I understand the problem better.

I added my comments to the PR.

plugin/slime.vim
thejohnfreeman: It could, but so can everything in the plugin. I think the purpose of a plugin is to capture common tasks and configurations and make them easily reusable. I think the 90% case is that a user wants to send keys to a pane in the same tmux session. This is that default. I'll remove it, but would it be fine if I moved it to the README as an example? At least then users can just copy-paste (which I believe is the second most common choice for initial configuration, after "no configuration, use the defaults").

plugin/slime.vim
thejohnfreeman: To explain `:.2`, I generally run vim in a tmux window, and then split the window into two panes when I want to start a REPL. The target `:.2` sends text to the other pane in the same window, to the REPL. The default target `:` sends the text to the pane running vim, and if I'm in normal mode, God help me.

I'll move this to the README too.

thejohnfreeman:
I missed the `slime_default_config` setting before. I'll use that in the README instead. Another change incoming.

jpalardy:
Looking good, thanks.

thejohnfreeman:
Thank you!

PR #130, opened 2017-03-03
Haskell slime escape text -

lynnard:
Allow import lines to be sent to ghci without being wrapped in blocks.

Currently when import lines are selected and sent over, they are sent in a multi-line block, and this causes ghci error as it only supports importing single modules at a time; the other problem is that these lines shouldn't be indented.


plugin/slime.vim
```diff
@@ -77,7 +77,9 @@ function! s:TmuxConfig() abort
   if !exists("b:slime_config")
     let b:slime_config = {"socket_name": "default", "target_pane": ":"}
   end
-  let b:slime_config["socket_name"] = input("tmux socket name or absolute path: ", b:slime_config["socket_name"])
+  if !(exists("g:slime_dont_ask_tmux_socket_name") && g:slime_dont_ask_tmux_socket_name)
```
jpalardy: interesting…

jpalardy:
I'm not in a position to evaluate the Haskell part -- because I don't use vim-slime w/ Haskell.

Sometimes I wish the "language-specific" were external scripts with their own unit tests :-)

jpalardy:
@dbeecham @clarkenciel @romeovs @lpil thoughts or comments?

plugin/slime.vim
lynnard: This was not supposed to be in this PR - I just pushed some unrelated change and ended up here :P Will take this out

jpalardy: I'm thinking about changing that code to be able to provide a function pointer to do the configuration.

It's a part of the code where everybody seems to have opinions about what's a reasonable default.

dbeecham:
This commit is unfortunately broken on literate haskell files; `Remove_initial_gt` should probably be done before `FilterImportLines(l:lines)`.

This, for example, does not work (test.lhs)

    > import Data.List
    > import Control.Monad

For some reason, a single line import works. Not sure why.

Other then that, code looks solid. Tried the most common use cases and a couple of more ugly cases, and the commit handles them fine. I say accept pull request - even if the commit is slightly broken, it's less broken then the current state, and code quality is OK.

@lynnard, can you see if you can fix the literate haskell issue as well? I think you just need to `Remove_initial_gt` first, or change your regex.

lynnard:
@dbeecham Thanks for testing - I've updated with a new commit now

dbeecham:
Nope! Undefined reference to `l:nonImports`. :) This should work:

    [...]
    let l:lines = Remove_initial_gt(l:lines)
    let [l:imports, l:nonImports] = FilterImportLines(l:lines)                                                                                             
    let l:lines = Remove_line_comments(l:nonImports)
    [...]

lynnard:
I've obviously forgot the importance of testing in a scripting language :) Thanks for the quick spot; this is updated now

jpalardy:
Excellent.

Thanks for the update, @dbeecham @lynnard

Issue #23, opened 2012-08-06
"Slurped X characters into buffer" message


idbrii:
Every time I trigger a send to slime, my screen session gives a message: "Slurped X characters into buffer." This message stays on the screen for several seconds before anything happens. Is it possible to disable it?

I found a similar issue on another screen plugin, but I can't figure out how to apply it to slime:

https://github.com/ervandew/screen/issues/4

travis-rodman:
Thanks for writing up the code modification @idbrii, I was having the same issue, helped out a lot.

travis-rodman:
actually, that borked my send after the second call... apparently is set a global variable and only the original command was sent again :/

PR #133, opened 2017-03-27
Erase the message issued by paste command immediately. No side-effect.

chiku-samugari:
As it was mentioned in #23, current approach to avoid the "Slurp X characters ..." message changes the `msgwait` set by the user. The proposed one erases that message immediately without such side effects.


jpalardy:
Looks good -- thanks :-)

jpalardy:
I had to fix the ^M (7d0374935e7bd4a3a3c2c5d7ac7d70c2650b89ae) … not sure what happened.

Can you confirm it still works for you?

chiku-samugari:
Yes, it works completely for me. FYI, `\n` and `\r` also works. But I think that the real `^M` is the most portable choice. Only the problem is that the character is a bit hard to see in some pagers :-)

Issue #136, opened 2017-05-12
[Haskell] Unwanted `let` inserted when typeclass declaration contains typeclass constraint


ethagnawl:
Sending the following to GHCi results in a parse error:

```
-- vim
instance Eq a => Eq (Identity a) where
  (==) (Identity v) (Identity v') = v == v'

-- GHCi
Prelude> :{
Prelude| let instance Eq a => Eq (Identity a) where
Prelude|       (==) (Identity v) (Identity v') = v == v'
Prelude| :}
<interactive>:0:0: parse error on input `instance'
```
For some reason, the typeclass constraint is causing the `let` to be introduced. A similar (albeit contrived) typeclass declaration _without_ a typeclass constraint works as expected:
```
-- vim 
instance Eq (Identity a) where
  (==) (Identity v) (Identity v') = False

-- GHCi
Prelude> :{
Prelude| instance Eq (Identity a) where
Prelude|       (==) (Identity v) (Identity v') = False
Prelude| :}
```

I'll take a pass it implementing a fix when I find a few minutes.

jpalardy:
My own contribution to that file is minimal. Feel free to ping someone from the [blame](https://github.com/jpalardy/vim-slime/blame/master/ftplugin/haskell/slime.vim)
if needed.

ethagnawl:
Closed via #137.

PR #137, opened 2017-05-29
136 - prevent let from being inserted before instance declaration using typeclass constraint

ethagnawl:
Addresses #136.

I'm open to suggestions about cleaner, more self-documenting ways of codifying this _should prepend let_ logic.


jpalardy:
You're the Haskell expert :-)

Issue #139, opened 2017-05-29
Incorrect newline behavior with Neovim target


NovaDev94:
### Expected Behavior

* `<Plug>SlimeLineSend` will send the current line with a single `newline` at the end.  

* `:SlimeSend0 <args>` will send the arguments without any `newline` at the end.  

* `:SlimeSend1 <q-args>` will send the arguments with a single `newline` at the end.  

### Actual Behavior (current)

- [x] `<Plug>SlimeLineSend` send the current line with a single `newline` at the end.  

- [ ] `:SlimeSend0 <args>` send the arguments WITH 1 `newline` at the end.  

- [ ] `:SlimeSend1 <q-args>` send the arguments WITH 2 `newline` at the end.

PR #140, opened 2017-05-29
Correct newline behavior for Neovim target

NovaDev94:
Resolve: https://github.com/jpalardy/vim-slime/issues/139


jpalardy:
It works if you say it works :-)

I'll ping you if I hear otherwise.

PR #147, opened 2017-07-22
Visual Mode Preserve Cursor position after long selections

ddxv:
This option allows users to select several lines of text using visual mode and after sending to vim-slime, preserve their cursor position at the bottom of the selection. 

Keeping the cursor position after Visual mode selection is accessed by unsetting `g:slime_preserve_curpos_sel` 

`let g:slime_preserve_curpos_sel = 0`

Note, unsetting `g:slime_preserve_curpos_sel` changes a default behaviour. When selecting a small portion of text in middle of a long single line of text using Visual mode the cursor will move to the beginning of line, no longer preserving it's cursor location in the middle of line.

### DIFF
--- a/README.md
+++ b/README.md
@@ -201,7 +201,7 @@ By default, vim-slime will try to restore your cursor position after it runs. If
 
     let g:slime_preserve_curpos = 0
 
-If you would like vim-slime to restore your cursor position after using Visual Mode in Vim, unset the `g:slime_preserve_curpos_sel` option:
+When selecting in Visual mode, if you would like vim-slime to restore your cursor position to the end of the selection, unset the `g:slime_preserve_curpos_sel` option:
 
     let g:slime_preserve_curpos_sel = 0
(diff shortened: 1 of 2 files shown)


jpalardy:
Hi there,

I see where you're going with this. But since you could press `']` after running vim-slime, isn't this a "key mapping" problem?

ddxv:
Hello,

Yes, I added '] to the end of the copy and send that vim-slime does as I
found it felt like a continuity break having the cursor move back to the
top of every selection. I often work with data and usually code from the
top down, often going over multiple cleaning steps over and over as I get
"farther" down each time before corrupting the data and starting over.
Perhaps it is not for everyone, I just found myself constantly wondering
why it always jumped the cursor back to the top when I had just brought it
down and eventually decided to fix it.

If it's not for vim-slime, no worries, I have my fork and fix.

Thanks,
James

PR #150, opened 2017-09-20
Guard missing keys in b:slime_config for all s:XXXXConfig functions,

randomizedthinking:
Now user can update 'g:slime_target' in one session, then use `C-C v` to configure the destination.

### DIFF
--- a/plugin/slime.vim
+++ b/plugin/slime.vim
@@ -41,8 +41,9 @@ function! s:ScreenConfig() abort
   if !exists("b:slime_config")
     let b:slime_config = {"sessionname": "", "windowname": "0"}
   end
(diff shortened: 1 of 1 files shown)


jpalardy:
I had a look and I'll have to get back to you on this.

For now, it duplicates the content of default configs (it's not DRY).

I also don't think it addresses the problem fully. There are some tests I want to run.

jpalardy:
What about [this](https://github.com/jpalardy/vim-slime/blob/65201427e125f8dc29ae434b7897e1eb854276ab/plugin/slime.vim#L178-L181)?

That's what I had in mind. If you use slime in a buffer, switch the `g:slime_target`, it won't query again for that target's specific config…

The code assume the existence of `b:slime_config` is enough evidence that it's configured correctly. But it won't be between targets.

jpalardy:
no reply, closing.

PR #152, opened 2017-09-25
Make SlimeSend preserve cursor position

### LINKED ISSUE #151: SlimeSend doesn't preserve cursor position
In a buffer like this:
```
1 blah # the first column is line number
2 blah
3 blah <- current cursor position
4 blah
```
and the cursor is at line 3, execute `:1,4 SlimeSend` will place the cursor at the beginning of line 1, expectation: cursor position should still be at line 3.

qwfy:
This resolves #151: SlimeSend doesn't preserve cursor position.


jpalardy:
Looking good, thanks.

PR #153, opened 2017-10-31
Support vim :terminal

gbarta:
Hi, I have added a target for the recently implemented vim :terminal command.  It is handled separately from neovim's implementation of :terminal, which is handled by the already existing 'neovim' target.

SlimeConfig allows selection of an existing running :terminal instance, or on-demand creation of a new :terminal instance.


jpalardy:
It's looking good.

Thanks :-)

PR #154, opened 2017-11-10
Add NeoVim terminal information to README.md

SirSpoony:
I was a little confused until I did some digging, this will help other newbies


jpalardy:
Whoah… I didn't know it was missing.

Thanks for the PR!

Issue #155, opened 2017-11-19
Add configuration for vim terminal


addadi:
it would be useful to have optional configuration for vim 8 terminal, for example:
let g:slime_default_config = {"vertical" : 1, "term_finish": "close"} 

and then in VimterminalDescription something like:
let new_bufnr = term_start(cmd, b:vimterminal_config)
(assuming vimterminal_config reads from slime_default_config)

Thanks,

jpalardy:
Hi there,

If you're willing to prototype this and send me a PR, I would certainly consider it.

addadi:
ok, I just did so (with my very humble viml skills)

jpalardy:
moved to #157

addadi:
you can close this now

jpalardy:
Thanks for the PR :-)

PR #156, opened 2017-11-20
prototype vim 8 terminal optional configuration

addadi:
to address issue #155  added support in optional configuration for vim 8 terminal, via let g:slime_default_config in vimrc

### DIFF
--- a/plugin/slime.vim
+++ b/plugin/slime.vim
@@ -172,12 +172,13 @@ function! s:VimterminalConfig() abort
     echoerr "vimterminal support requires vim built with :terminal support"
     return
   endif
-  let b:vimterminal_config =  {}
   if !exists("b:slime_config")
       let b:slime_config = {"bufnr": ""}
-  else
-      let b:vimterminal_config = b:slime_config
-  endif
+  end
+  let b:vimterminal_config =  {}
+  if exists("g:slime_vimterminal_config")
+    let b:vimterminal_config  = g:slime_vimterminal_config
+  end
   let bufs = filter(term_list(),"term_getstatus(v:val)=~'running'")
   let terms = map(bufs,"getbufinfo(v:val)[0]")
   let choices = map(copy(terms),"s:VimterminalDescription(v:key+1,v:val)")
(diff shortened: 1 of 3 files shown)


plugin/slime.vim
```diff
@@ -172,9 +172,12 @@ function! s:VimterminalConfig() abort
...
     return
   endif
+  let b:vimterminal_config =  {}
   if !exists("b:slime_config")
-    let b:slime_config = {"bufnr": ""}
-  end
+      let b:slime_config = {"bufnr": ""}
+  else
+      let b:vimterminal_config = b:slime_config
+  endif
```
jpalardy: not what the advantage of doing this here ^

see below

plugin/slime.vim
```diff
@@ -189,7 +192,7 @@ function! s:VimterminalConfig() abort
         let cmd = &shell
       endif
       let winid = win_getid()
-      let new_bufnr = term_start(cmd)
+      let new_bufnr = term_start(cmd, b:vimterminal_config)
```
jpalardy: would this be better as a new variable? `g:slime_vimterminal_config` which you could fill with anything?

You would have to handle missing or empty cases too.

addadi: ok, I've created a new variable and handled empty cases as well. I've also updated the documentation to reflect this

jpalardy: Why both a `b:` and a `g:` variables? Would there be a case where you would want buffer-specific overrides?
In practice, what would that look like?

jpalardy: If you don't expect `b:`-level overrides, would this be simpler?

```
if exists("g:slime_vimterminal_config")
  let new_bufnr = term_start(cmd, g:slime_vimterminal_config)
else
  let new_bufnr = term_start(cmd)
end
```

This would go around line 195. And line 176-180 wouldn't have to be touched.

addadi: your suggestion make sense. should I push a new commit with the correction?

jpalardy:
I updated the code in a separate PR (#157).

Let's go there and finish this.

PR #157, opened 2017-11-23
addadi's master, plus tweaks

jpalardy:


doc/vim-slime.txt
```diff
@@ -150,6 +150,12 @@ your |.vimrc|:
 >
         let g:slime_target = "vimterminal"
 <
+Vim terminal configuration can be set by using the following in your .vimrc:
+
+    let g:slime_vimterminal_config = {options}
```
jpalardy: Maybe there could be an example here ^

`let g:slime_vimterminal_config = {"vertical" : 1, "term_finish": "close"}`

addadi:
great! thanks!

PR #170, opened 2018-10-02
Tmux bracketed mode paste enabled

randomizedthinking:
And the possible ending CR is treated separately given its special meaning for REPL.

It solved the indentation issues for tmux. I checked now no magic is required to paste to python sessions under tmux.

The code is not verified under windows.

### DIFF
--- a/plugin/slime.vim
+++ b/plugin/slime.vim
@@ -60,9 +60,18 @@ function! s:TmuxCommand(config, args)
 endfunction
 
 function! s:TmuxSend(config, text)
-  call s:WritePasteFile(a:text)
+  let l:lastchar = a:text[len(a:text)-1]
+  if l:lastchar == "\n"
+      call s:WritePasteFile(a:text[0:(len(a:text)-2)])
+  elseif
+      call s:WritePasteFile(a:text)
+  endif
   call s:TmuxCommand(a:config, "load-buffer " . g:slime_paste_file)
-  call s:TmuxCommand(a:config, "paste-buffer -d -t " . shellescape(a:config["target_pane"]))
+  call s:TmuxCommand(a:config, "paste-buffer -p -d -t " . shellescape(a:config["target_pane"]))
+  if l:lastchar == "\n"
+      call s:TmuxCommand(a:config, "set-buffer ")
+      call s:TmuxCommand(a:config, "paste-buffer -d -t " . shellescape(a:config["target_pane"]))
+  endif
 endfunction
 
 function! s:TmuxPaneNames(A,L,P)


randomizedthinking:
I am not an expert on the topic either. Here is a link at Xterm Control Sequence. Again, quite thin.

> [Xterm Control Sequences/Bracketed paste mode](http://xfree86.org/current/ctlseqs.html#Bracketed%20Paste%20Mode)

My understanding is that bracketed paste mode is now default for xterm, rxvt, vim, emacs. Before, we need to `:set paste` in vim for pasting text, but no longer needed now.

jpalardy:
This will solve the problem when the user is using a combination of `tmux` + `python`?

Are there other combinations that I need to be aware of?

randomizedthinking:
Basically, it solves the problem when vim-slime target is a tmux pane.
Personally, I tested python/ipython, psql, snowsql, sas -- they all work
fine.

Also, the implementation has no negative side-effect. If the receiving end
does not support bracketed paste mode, it simply falls back to the same
behavior as the current implementation.

jpalardy:
Sounds good.

Does it make the [python ftplugin](https://github.com/jpalardy/vim-slime/blob/master/ftplugin/python/slime.vim) redundant? Does the documentation need to be updated?

randomizedthinking:
Now `python ftplugin` is only redundant for the `tmux` target. Would make sense to update documentation for the tmux section.

ajweiss:
I hacked up my local copy to do this quite a while ago and it has been working well, however I was hesitant to write up a PR as I wasn't sure if there were other REPLs that would behave differently or break when they see bracketed paste control characters from tmux.  Older copies of ipython were also a concern, it seems that they've been changing this behavior over time.

Also, in my copy I wrote out the final newline to the paste file and did another load/paste rather than using set-buffer with a bare ^M.  They're both sort of ugly workarounds, but hidden control characters can be confusing and it has the nice side effect of clearing the paste file.

```
function! s:TmuxSend(config, text)
  call s:WritePasteFile(a:text)
  call s:TmuxCommand(a:config, "paste-buffer -p -d -t " . shellescape(a:config["target_pane"]))
  call system("cat > " . g:slime_paste_file, "\n")
  call s:TmuxCommand(a:config, "paste-buffer -d -t " . shellescape(a:config["target_pane"]))
endfunction
```

@jpalardy I have some other tweaks laying around where I use mktemp from coreutils to create/name the temp file and automatically wipe temp files after sending if the user doesn't supply an explicit filename.  Are you interested in those?

jpalardy:
@ajweiss Not being a ipython user myself, I'm not sure I want to merge this: it's a specific fix for `ipython + tmux`.

I can see this living on a branch, and point users there if they want to experiment?

ajweiss:
I agree.  That was my hesitation!  How would you feel about some sort of mechanism for allowing language specific tweaks to live in the ftplugins that are accessible in the Send() functions?  Then these weird one-offs that are an unavoidable consequence of gluing together external software with all its quirks could be included to keep things working without making too much of a mess...

Maybe generalize EscapeText to return both the escaped text and a vimscript dictionary of flags, where one would be FlagTmuxUseBracketedPaste or somesuch...  or have a separate dispatch to separate flags functions (although being able to set flags while examining the text may be useful).

ajweiss:
It seems there should be some way to configure it in master though.  Not only does it work around the apparent bug in IPython 7.x, but it also seems to be "the right way" to be pasting into terminals.  IPython picks it up from prompt_toolkit which recently added it and readline added support in 7.x.  tmux supports it and only uses it if the client program sends escape codes asking for it to be turned on.  (which is no guarantee it will work, but at least it makes no difference by default)

Also with respect to the python ftplugin, it only makes it redundant when pasting to IPython or Jupyter Console which will do cleanup if bracketed pasting is on (or via the %cpaste and %paste magics).  Plain old CPython doesn't attempt any cleanup and will complain regardless if indentation isn't right.

jpalardy:
Going back over the code:

* if newline needs to be sent separately, that logic can be done in the [python ftplugin](https://github.com/jpalardy/vim-slime/blob/0c5fca7698f59848967702ff96f5b914ffc3fefb/ftplugin/python/slime.vim)
* then `TmuxSend` doesn't need to change beyond added the `-p` flag? (which wouldn't hurt)

How does that sound?

randomizedthinking:
@jpalardy I think it would be better to handle the newline in vimscript since the last newline is needed not only for ipython but for any REPLs under the bracketed paste mode.

The rationale is that if a newline is sent over to REPL through bracketed paste mode, the REPL just treats it like some regular text with no side effect -- that's why the indentation is kept yet the statement won't be executed. So a final newline under normal paste mode is expected to trigger the side effect.

oblitum:
This pull fixes #191.

oblitum:
I also tried [ajweiss](https://github.com/jpalardy/vim-slime/pull/170#issuecomment-437723892) and it also works to fix that, and it's also cleaner.

jpalardy:
Hi @oblitum 

I didn't close #170 ... but I was hesitant to patch the tmux logic if:

- the fix is specific (just python)
- the fix is general (not just tmux)

If you'd like to take another look and come up with a proposal, I'll have a look.

oblitum:
>     * the fix is specific (just python)

It's not, as it fixes an unrelated program, pgcli.

>     * the fix is general (not just tmux)

It's not, as it only adds 3 lines to `s:TmuxSend` specifically, to take advantage of tmux's bracketed paste mode support.

> If you'd like to take another look and come up with a proposal, I'll have a look.

I can surely send a pull picking the [3 lines from ajweiss's patch](https://github.com/ajweiss/vim-slime/commit/68cf5ca5196fe293968de6f32234be16fc3e2d55). But I don't having anything more to add myself to the matter.

jpalardy:
Let's do that, send up the PR and I'll merge it.

oblitum:
@jpalardy #205.

PR #171, opened 2018-10-03
Fix #161 extra newline with vim8 terminal

### LINKED ISSUE #161: extra newline when sending to vim8 :terminal?
I'm using `g:slime_target = 'vimterminal'`, most recent vim and vim-slime, archlinux, zsh/bash.  I've noticed `:SlimeSendLine` sends an extra `<cr>` to the terminal, in addition to the one that runs the command.  It also happens with `<plug>SlimeRegionSend` when selecting a partial line, so I think it's not a matter of e.g. accidentally grabbing the current line's newline.  vim-slime seems to be adding two newlines when the sent text does not have one already.

As an example, where `[ ]` is the visual selection,

    [some line] some extra

Pressing `<plug>SlimeRegionSend` in visual mode, the terminal would look like:

    $ some line<cr>
    $ <cr>
    $ [cursor]

`<cr>` for emphasis.

So, is this the expected behavior?  I can see how it might be intentional and useful to separate multiple terminal sends, but I didn't see anything about it in the documentation or existing issues mentioning it.

However, I'm wondering if it's possible to disable this?  I am running a command which expects keyboard input, and the newline is being sent into that interactive command.

gbarta:
Fix #161 extra newline with vim8 terminal

When sending each line with a carriage return, make sure that the number
of carriage returns is not more than the number of newlines in the text.

### DIFF
--- a/plugin/slime.vim
+++ b/plugin/slime.vim
@@ -153,8 +153,8 @@ function! s:VimterminalSend(config, text)
   " Ideally we ought to be able to use a single term_sendkeys call however as
   " of vim 8.0.1203 doing so can cause terminal display issues for longer
   " selections of text.
-  for l in split(a:text,"\n",1)
-    call term_sendkeys(bufnr,l."\<cr>")
+  for l in split(a:text,'\n\zs')
+    call term_sendkeys(bufnr,substitute(l,'\n',"\r",''))
     call term_wait(bufnr)
   endfor
 endfunction


plugin/slime.vim
```diff
@@ -153,8 +153,8 @@ function! s:VimterminalSend(config, text)
   " Ideally we ought to be able to use a single term_sendkeys call however as
   " of vim 8.0.1203 doing so can cause terminal display issues for longer
   " selections of text.
-  for l in split(a:text,"\n",1)
-    call term_sendkeys(bufnr,l."\<cr>")
+  for l in split(a:text,'\n\zs')
```
jpalardy: Vim help:

> \zs     Matches at any position, and sets the start of the match there: The
        next char is the first char of the whole match.

What does it mean for that `split`?

plugin/slime.vim
```diff
@@ -153,8 +153,8 @@ function! s:VimterminalSend(config, text)
...
+    call term_sendkeys(bufnr,substitute(l,'\n',"\r",''))
```
jpalardy: Replace the first `\n` with `\r`?

jpalardy:
I'm trying to give my future self a helping hand :-)

plugin/slime.vim
gbarta: split() typically splits text at the regex match, and also removes the matched text.  The \zs modifies the match to start at that point (the end in this case) and thus have zero length so nothing is removed.  The match position is also after the \n so that the effect is that each component of the split will have a \n at the end of it except possibly the very last one which may or may not have a \n at the end depending on the original text.
I agree that the vim help for \zs doesn't make this especially clear, but the help for split() explicitly mentions this usage, which where I discovered it.

plugin/slime.vim
gbarta: Yes.  The flag could have 'g' since we want to replace all \n with \r, but the effect of the split above is that there is never more than 1 \n in the string so the 'g' would have no effect.

plugin/slime.vim
jpalardy: Nice!

jpalardy:
Ok, it's looking good 👍

PR #173, opened 2018-10-30
cleaner support for matlab .m files

mcolligan:
Hello,

I use vim-slime with MATLAB. When sending text that includes a tab character, the command window spits out "No completions found." This can make reading that window difficult when littered with one of those for every tab that goes by. These changes replace the tab character with 4 spaces and stops this behaviour. It also removes comment lines which are not needed.

The code actually all came from the haskell ftplugin file with only minimal changes needed to get it to work.

Maybe this might be useful if anybody else uses vim-slime and MATLAB!


jpalardy:
Looking good, thanks!

PR #175, opened 2018-11-12
Use mktemp(1) for paste files and auto-remove them by default

ajweiss:
If no explicit paste file is specified, use mktemp(1) to generate paste files and wipe them when finished to stay more in line with standard UNIX tmpfile practices and avoid potential security problems involving tmpfiles and symlinks.  mktemp(1) is standard on UNIX/MacOS.

This won't work on Windows unless there's Cygwin/MKS/etc installed, but paste files don't appear to be used on Windows now anyway.


wincent:
How is this better than just using `tempname()` as suggested in the docs, which avoids shelling out? I only tested this on macOS in Neovim and Vim, but with both `tempname()` generates a random-gibberish path like "/var/folders/tl/04jnvr4s34v704k5h2rqjmkh0000gn/T/vEye0ji/3" that gets auto-cleaned-up by the system and is in a location only readable by the current user.

ajweiss:
I initially implemented this with vim's `tempname()` but backed away from it for some reason (I was initially attracted to it because it would work on Windows).  I did this a few months ago so I don't remember exactly why, but let me see if I can refresh my memory.

ajweiss:
I think I was spooked by this:

```
The result is a String, which is the name of a file that
doesn't exist.  It can be used for a temporary file.  The name
is different for at least 26 consecutive calls.  Example:
```

Combine this weird "different for 26 consecutive calls" business with the fact that even though the directory is protected, the function does not create the tmpfile (and they're easily guessable), I decided to better be safe than sorry and use the standard facility.

wincent:
I think those docs are stale, probably reflecting a time when it created letter-based names. If I run `tempname()` bunch of times I see it using numbers, presumably without any practical limit:

```
/var/folders/tl/04jnvr4s34v704k5h2rqjmkh0000gn/T/v39Wd7Z/17
/var/folders/tl/04jnvr4s34v704k5h2rqjmkh0000gn/T/v39Wd7Z/18
/var/folders/tl/04jnvr4s34v704k5h2rqjmkh0000gn/T/v39Wd7Z/19
/var/folders/tl/04jnvr4s34v704k5h2rqjmkh0000gn/T/v39Wd7Z/20
/var/folders/tl/04jnvr4s34v704k5h2rqjmkh0000gn/T/v39Wd7Z/21
/var/folders/tl/04jnvr4s34v704k5h2rqjmkh0000gn/T/v39Wd7Z/22
/var/folders/tl/04jnvr4s34v704k5h2rqjmkh0000gn/T/v39Wd7Z/23
/var/folders/tl/04jnvr4s34v704k5h2rqjmkh0000gn/T/v39Wd7Z/24
/var/folders/tl/04jnvr4s34v704k5h2rqjmkh0000gn/T/v39Wd7Z/25
/var/folders/tl/04jnvr4s34v704k5h2rqjmkh0000gn/T/v39Wd7Z/26
/var/folders/tl/04jnvr4s34v704k5h2rqjmkh0000gn/T/v39Wd7Z/27
/var/folders/tl/04jnvr4s34v704k5h2rqjmkh0000gn/T/v39Wd7Z/28
/var/folders/tl/04jnvr4s34v704k5h2rqjmkh0000gn/T/v39Wd7Z/29
/var/folders/tl/04jnvr4s34v704k5h2rqjmkh0000gn/T/v39Wd7Z/30
/var/folders/tl/04jnvr4s34v704k5h2rqjmkh0000gn/T/v39Wd7Z/31
/var/folders/tl/04jnvr4s34v704k5h2rqjmkh0000gn/T/v39Wd7Z/32
```

jpalardy:
I'm also not in love with the idea of shelling out.  :-/

ajweiss:
Yeah, it is a little janky.  I tried running it in a loop for thousands of iterations and it seems fine.  I still don't remember what my original reasoning was for abandoning `tempname()` other than maybe smell?  (weird comments in docs, use of predictable repeating integer)  I must have been pretty convinced as I wasn't happy about the dependency on platform.

That said, I suppose if the higher level folder is owned and mode 700 by the user who started the process, there's not much risk.

Happy to revert it to `tempname()` if you guys think it's the way to go...

Issue #166, opened 2018-06-25
"let" assigment on ghci is no longer needed with ghci-8.2+


mimi1vx:
https://github.com/jpalardy/vim-slime/blob/e58486b791e6e1126ffd46d4ef634b8fef104f7f/ftplugin/haskell/slime.vim#L19

jpalardy:
no reply

jpalardy:
@mickk-on-cpp 

Would you be open to?
- creating a PR
- making this configurable (so people can switch between modes)
- update the Haskell-related documentation

PR #177, opened 2018-11-29
Add Haskell preprocessing options.

mickk-on-cpp:
As suggested in #166, add one option to control the preprocessing of Haskell top-level bindings into a let binding group. This preprocessing is on by default, i.e. this doesn't change from the current behaviour.

I tend to have some reservations with such a piecemeal approach. One way or the other the current handling of Haskell is rough around the edges and could benefit from a more comprehensive, less myopic approach. (But of course that would take time, and I have too much on my plate myself right now.)

Tested (*very* casually) with GHC 7.10.3 and 8.4.4.

### DIFF
--- a/ftplugin/haskell/README.md
+++ b/ftplugin/haskell/README.md
@@ -1,30 +1,41 @@
 
 ### Haskell
 
-This plugin has support for sending haskell source code to the `ghci`. Syntax differences between `ghci`
-are automatically detected and fixed and comments (which aren't allowed in `ghci`) are filtered. Try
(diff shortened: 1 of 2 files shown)


jpalardy:
looking good

jpalardy:
Another more general approach would be to provide command-line filters, specified by configuration. For example, if you had a script to "fix" Haskell code named "hs-fix" -- you could set it up to filter text before feeding it to tmux.

```
vim --> command-line filters (0 to many) --> tmux --> whatever REPL
```

This would require revisiting how vim-slime is currently setup.

Issue #178, opened 2018-12-03
Wrong Tmux panel indexing instructions in doc?


oblitum:
Hi, just trying this plugin, it's great. I'm using tmux, and it seems docs are not correct at `":i"`, it should be `":.i"`, at least this is what worked for me (tried with `":.2"` after setting it up again with <kbd>ctrl</kbd><kbd>c</kbd>, <kbd>v</kbd>, my tmux indexing starts at 1). I figured that out from this line which I'm not using `let g:slime_default_config = {"socket_name": split($TMUX, ",")[0], "target_pane": ":.2"}`.

jpalardy:
Hi there,

> it's great

Thanks :-)

Can you explain your setup and give me an example of what works and what doesn't? If possible, a PR would be appreciated too.

oblitum:
Hi, it's just that README and docs states to index tmux with `":i"` syntax, but what works for indexing panes has been `":.i"`, `:i`, without `.`, doesn't work.

oblitum:
This is my `tmux.conf` where I set indexing to 1, which is a common approach, but that's not particularly relevant, I just have to use ":.2" on a two sided split, instead of ":.1":

- https://github.com/oblitum/dotfiles/blob/ArchLinux/.tmux.conf#L17-L18

jpalardy:
Ok, here's what I tried:

```bash
> tmux ls
# ...snip...
20: 1 windows (created Tue Dec  4 09:10:05 2018) [160x44] (attached)
```
That session has 1 window, 1 pane.

```bash
> seq 3 14 > nums.txt
> tmux load-buffer nums.txt; tmux paste-buffer -d -t 20:
> tmux load-buffer nums.txt; tmux paste-buffer -d -t 20:0
> tmux load-buffer nums.txt; tmux paste-buffer -d -t 20:0.0
```

These all work the same ^. (for me)

From within the session, `:`, `:0`, `:0.0` also work.

oblitum:
```
~ ❯❯❯ tmux load-buffer nums.txt; tmux paste-buffer -d -t :2
3
4
5
6
7
8
9
10
11
12
13
14
~ ❯❯❯ 3
cd: no such entry in dir stack
~ ❯❯❯ 4
cd: no such entry in dir stack
~ ❯❯❯ 5
cd: no such entry in dir stack
~ ❯❯❯ 6
cd: no such entry in dir stack
~ ❯❯❯ 7
cd: no such entry in dir stack
[...]

PR #179, opened 2018-12-04
Mention tmux's special tokens for pane indexing

oblitum:
As per https://github.com/jpalardy/vim-slime/issues/178#issuecomment-444199477.

Notice that I've also slightly updated the original example. I can revert that and rebase. I did it for it to reflect what is in the [advanced section](https://github.com/jpalardy/vim-slime#advanced-configuration), which is simpler what I was using previously, not sure why it's more complex here, for the same effect.

### DIFF
--- a/README.md
+++ b/README.md
@@ -126,17 +126,24 @@ tmux target pane:
 
 Note that all of these ordinals are 0-indexed by default.
 
-    ":"     means current window, current pane (a reasonable default)
-    ":i"    means the ith window, current pane
-    ":i.j"  means the ith window, jth pane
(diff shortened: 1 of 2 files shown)


jpalardy:
awesome, thanks 👍

jpalardy:
I even updated my [own slime config](https://github.com/jpalardy/dotfiles/commit/7dac70922a7f566a9e58107ed97c0a9a91d8bb23) :-)

oblitum:
Nice. I was thinking whether it's a good plugin default or not.

Issue #181, opened 2018-12-21
please make g:slime_vimterminal_cmd


jDrunken:
I always use Node in vim-slime with macvim.
annoying to type command 'node', so make Global variable like `g:slime_vimterminal_cmd`

like this,

```
# ./plugin/slime.vim :: line 187

if !exists("g:slime_vimterminal_cmd")
    let cmd = input("Enter a command to run [".&shell."]:")
    if len(cmd)==0
        let cmd = &shell
    endif
else
    let cmd = g:slime_vimterminal_cmd
endif
```

thank you!

jpalardy:
Hi @jDrunken 

Send me a PR for this 😃 

Please include a comment in the README to document the flag.

jDrunken:
merged. thanx. :)

PR #182, opened 2018-12-24
add frequently used command option for vimterminal

jDrunken:
follow this.

https://github.com/jpalardy/vim-slime/issues/181#issuecomment-449261570

thanx & merry christmas :)


jpalardy:
looking good, thanks ^

PR #183, opened 2019-01-10
Move bulk of functionality into autoloaded file

wincent:
This should improve startup time. Here are the relevant excerpts from `vim vim --startuptime`, five runs, before and after.

Before:

    000.369: sourcing $HOME/.vim/plugin/slime.vim
    000.380: sourcing $HOME/.vim/plugin/slime.vim
    000.335: sourcing $HOME/.vim/plugin/slime.vim
    000.334: sourcing $HOME/.vim/plugin/slime.vim
    000.353: sourcing $HOME/.vim/plugin/slime.vim

After:

    000.141: sourcing $HOME/.vim/plugin/slime.vim
    000.138: sourcing $HOME/.vim/plugin/slime.vim
    000.127: sourcing $HOME/.vim/plugin/slime.vim
    000.127: sourcing $HOME/.vim/plugin/slime.vim
    000.125: sourcing $HOME/.vim/plugin/slime.vim


wincent:
Only lightly tested this, as I just installed vim-slime last night. Just wanted to see if you think it's directionally correct.

FWIW, there a bunch of functions in the `ftplugin` directory that should probably be autoloaded too. For example, both the Haskell and the Matlab ftplugins define a function called `Is_comment` which is global, so if you open a Haskell file then a Matlab one, the second version of the function will clobber the other. There are many other global functions defined with names that might clash with functions written by the user of provided by other plug-ins (eg. `Tab_to_spaces`, `Lines`, `Unlines` etc). Those should probably all be made unique and moved into files like:

```
autoload/slime/haskell.vim:
function! slime#haskell#is_comment() abort
  " ...
endfunction

autoload/slime/matlab.vim:
function! slime#haskell#is_comment() abort
  " ...
endfunction
```

Either that, or make them script-local inside the ftplugin files. (Which, now that I write that, probably makes the most sense.)

jpalardy:
Hi there,

Even though this is my plugin, I'm not a vimscript expert. If the content moves to `autoload`, what's the impact in practice? (when does it load, etc?)

But, yes, if you can make this work -- and extend the practice to the `ftplugin`, that sounds like good stuff.

wincent:
> what's the impact in practice? (when does it load, etc?)

The idea is that you put only mappings and commands in the `plugin` directory; these get eagerly loaded at startup. Everything else goes into the `autoload` directory, and only gets loaded on the first actual call. In this way, you avoid parsing and evaluating everything up front unnecessarily early. In general, you might only save a few milliseconds this way (depending on the size and complexity of the plug-in code), but when a user has dozens or as many as a hundred plug-ins, these slices of time add up to a noticeable delay — "death by a thousand cuts", as they say.

[More on autoloading here](http://learnvimscriptthehardway.stevelosh.com/chapters/53.html).

As far as the `ftplugin` stuff goes, they already benefit from lazy-loading (they are only evaluated when the corresponding file-type is detected), so I think most of the functions should just be turned into script-local ones (ie. with an `s:` prefix) and not actually autoloaded. The ones that would benefit from being autoloaded are the ones that are shared across multiple files; there may not be many of those, but at least `Tab_to_spaces` looks like it is copy-pasted at least once and so would benefit being switched to a single, shared, autoloaded implementation.

wincent:
So I created a separate PR to deal with the hygiene of the ftplugins w.r.t the global namespace: https://github.com/jpalardy/vim-slime/pull/184

It applies cleanly alongside this PR without merge conflicts because they're touching different files.

jpalardy:
Sounds good, I'll merge both and we'll take it from there.

PR #185, opened 2019-01-22
Add kitty support

mdzhang:
Adds support for [kitty](https://sw.kovidgoyal.net).


jpalardy:
looking good, thank you 😃

Issue #172, opened 2018-10-19
Paste issue in IPython 7.0.1


gruevyhat:
I recently upgraded IPython to 7.0.1, and for some reason vim-slime has stopped working. When I hit `<ctrl-c><ctrl-c>` in Vim, it successfully inserts `%cpaste` into IPython, but then just stops---it doesn't paste any of the actual python commands. If I hit `<ctrl-c>` within IPython, it suddenly pastes all the lines, but only one line at a time, not as a code block; finally it pastes `--`, prompting a syntax error. I checked to make sure .slime_paste contains the python content, and it does. I have `let g:slime_python_ipython = 1` set in .vimrc. I get the same behavior in Gnu screen and tmux. Any help would be greatly appreciated.

jpalardy:
Hi @gruevyhat 

You might want to look at the other [iPython issues](https://github.com/jpalardy/vim-slime/issues?utf8=%E2%9C%93&q=is%3Aissue+is%3Aopen+ipython) and [this PR](https://github.com/jpalardy/vim-slime/pull/170).

Also:
* what's your setup?
* did you try a barebone vim with only vim-slime?

jpalardy:
Damn.

Who wants to take a stab at this?

gmnimrod:
I am having the same problem.
My setup: IPython 7.2.0, neovim 0.2.2, tmux 2.6, vim-slime 06525ceee2457bed7c7718fef0b4fe715651c392.

I have a possible explanation.
The text that Slime prepares is copied to ipython before the %cpaste command is executed.
When %cpaste command is executed, ipython enters the special cpaste mode.
Slime already copied all the text, so ipython just hangs, waiting for input.
After breaking with ctrl-c, ipython resumes reading lines from the regular input stream.
The code is already there. Ipython executes the code line by line.

Another observation: When using slime to send a large number of lines (~1000 on my setup) the second half of the code is pasted in cpaste mode, and the first half is pasted after exiting cpaste mode. This behavior is consistent with the above explanation.

This issue seems related to #142.

Adding a delay in `slime#send`, as suggested by @fx-kirin in #142 solves this issue.
I think that it is enough to add the delay once, only after "%cpaste -q\n".
So, the solution might be something like this:

```vim
function! slime#send(text)
  call s:SlimeGetConfig()

  let pieces = s:_EscapeText(a:text)
  " begin change
  if exists('g:slime_python_ipython')
[...]

PR #186, opened 2019-01-27
Fix paste issues in IPython. Add g:slime_dispatch_pause global option.

gmnimrod:
**Summary of Changes**
1. slime#send accepts a mixed list of strings and integers as an argument. The strings are sent with SlimeDispatch. The integers trigger a delay. The length of the delay is the value of the integer in ms.
1. IPython paste issues are fixed by pausing for g:slime_dispatch_ipython_pause milliseconds after sending the "cpaste" command.

**Discussion**
1. Where do you think is the correct place to document the new variable (in the README and/or docs)? For now, it is undocumented.
1. The first commit implements the (second) solution discussed in #172. The second commit implements an improved version (I hope...). The change made enables defining a different delay for each need. Let me know what you think.

I am open to criticism, so feel free to comment. I will make changes as needed.

### DIFF
--- a/autoload/slime.vim
+++ b/autoload/slime.vim
@@ -25,7 +25,8 @@ function! s:ScreenSend(config, text)
         \ " -X eval \"readreg p " . g:slime_paste_file . "\"")
   call system("screen -S " . shellescape(a:config["sessionname"]) . " -p " . shellescape(a:config["windowname"]) .
         \ " -X paste p")
-  call system('screen -X colon ""')
+  call system('screen -X colon "
+"')
 endfunction
 
 function! s:ScreenSessionNames(A,L,P)
@@ -347,7 +348,7 @@ function! slime#send(text)
   " will flush the rest of the buffer given a special sequence (ctrl-v)
   " so we, possibly, send many strings -- but probably just one
   let pieces = s:_EscapeText(a:text)
-  for piece in pieces
+    if type(piece) == 0   " a number
     if type(piece) == 0
       if piece != 0
         execute 'sleep' piece . 'm'
--- a/autoload/slime.vim
+++ b/autoload/slime.vim
@@ -15,10 +15,6 @@ if !exists("g:slime_paste_file")
   let g:slime_paste_file = expand("$HOME/.slime_paste")
 end
 
-if !exists("g:slime_dispatch_pause")
-  let g:slime_dispatch_pause = 100
-end
-
 """""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""
 " Screen
 """""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""
@@ -352,8 +348,8 @@ function! slime#send(text)
   " so we, possibly, send many strings -- but probably just one
   let pieces = s:_EscapeText(a:text)
   for piece in pieces
-    if piece == v:null
-      execute 'sleep' g:slime_dispatch_pause . 'm'
+    if type(piece) == 0
+      execute 'sleep' piece . 'm'
     else
       call s:SlimeDispatch('Send', b:slime_config, piece)
     end
--- a/autoload/slime.vim
+++ b/autoload/slime.vim
@@ -348,8 +348,8 @@ function! slime#send(text)
   " so we, possibly, send many strings -- but probably just one
   let pieces = s:_EscapeText(a:text)
   for piece in pieces
-    if type(piece) == 0
-      if piece != 0
+    if type(piece) == 0  " a number
+      if piece > 0  " sleep accepts only positive count
         execute 'sleep' piece . 'm'
       endif
     else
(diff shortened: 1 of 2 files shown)


autoload/slime.vim
```diff
@@ -348,7 +348,13 @@ function! slime#send(text)
   " so we, possibly, send many strings -- but probably just one
   let pieces = s:_EscapeText(a:text)
   for piece in pieces
-    call s:SlimeDispatch('Send', b:slime_config, piece)
+    if type(piece) == 0
```
jpalardy:
```suggestion
    if type(piece) == 0   " a number
```

jpalardy: might be better with a comment

autoload/slime.vim
```diff
@@ -348,7 +348,13 @@ function! slime#send(text)
...
+      if piece != 0
```
jpalardy: would that be better as `piece > 0`?   🤔

jpalardy: yeah… there's no "great" way to flag these things

I think it's ok to ignore negative and zero values.

autoload/slime.vim
```diff
@@ -347,8 +348,14 @@ function! slime#send(text)
   " will flush the rest of the buffer given a special sequence (ctrl-v)
   " so we, possibly, send many strings -- but probably just one
   let pieces = s:_EscapeText(a:text)
-  for piece in pieces
-    call s:SlimeDispatch('Send', b:slime_config, piece)
+    if type(piece) == 0   " a number
```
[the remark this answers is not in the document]
gmnimrod: Sorry :flushed: my bad
I tried using the "commit suggestion" button on github. Something went wrong.
I'll revert the commit.

autoload/slime.vim
```diff
@@ -25,7 +25,8 @@ function! s:ScreenSend(config, text)
         \ " -X eval \"readreg p " . g:slime_paste_file . "\"")
   call system("screen -S " . shellescape(a:config["sessionname"]) . " -p " . shellescape(a:config["windowname"]) .
         \ " -X paste p")
-  call system('screen -X colon ""')
+  call system('screen -X colon "
```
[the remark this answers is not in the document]
gmnimrod: Sorry :flushed: my bad
I tried using the "commit suggestion" button on github. Something went wrong.
I'll revert the commit.

autoload/slime.vim
gmnimrod: OK, I'll change it.

gmnimrod:
Should be fine now.
Do you want me to add `g:slime_dispatch_ipython_pause` to the docs?

RhysU:
How did you pick the default pause setting of 100?  Is the result pleasant for pasting, say, 30 lines?

gmnimrod:
> How did you pick the default pause setting of 100? Is the result pleasant for pasting, say, 30 lines?

Good question.
I picked it because it works on my setup. I figured it's a long amount of time for IPython, but a short amount of time for a human being.

There is only one pause on every time `%cpaste` is used. Slime is smart about it, so I think it should be fine. For example:

- If you call `30<Plug>SlimeLineSend`, there will only be a 100 ms delay, because slime sends all the lines together in one `%cpaste` operation, and not one by one.
- If you call `<Plug>SlimeLineSend` 30 times in a row separately, there will be no delay, because `%cpaste` is only used when sending more then 1 line.
- If you call `<Plug>SlimeParagraphSend` on a paragraph with 30 lines there will only be a 100 ms delay.
- If you call `<Plug>SlimeParagraphSend` 30 times in a row separately, there will be a total delay of 3 sec.

Only the last example is problematic, but I don't think it is very common.
What do you think?

RhysU:
That sounds very sensible. Thank you for explaining. And for fixing this Ipython issue. I ran into it yesterday and was thrilled that it was already reported with a fix under review.

jpalardy:
Looking good, thank you 👍 

I don't use ipython, you're going to see if it behaves well. If we need to tweak things, let me know.

neddles:
I greatly appreciate this effort. However, I am getting funny behavior. It appears to work fine if there is a trailing newline in the selected text. If there is no newline then "--" (as required by %cpaste) is shown, but the code does not execute in IPython. I have added an additional newline character here `vim-slime\ftplugin\python\slime.vim`:

-        `return ["%cpaste -q\n", g:slime_dispatch_ipython_pause, a:text, "--\n"]`
+        `return ["%cpaste -q\n", g:slime_dispatch_ipython_pause, a:text, "\n--\n"]`

but I don't think this is a robust solution. Perhaps the `g:slime_dispatch_ipython_pause` isn't executing properly?

neddles:
Actually... that doesn't work either. The same lines of code will sometimes execute in IPython, and sometimes they won't, within the same session. Seems like a timing issue to me, but increasing `g:slime_dispatch_ipython_pause` to 1 second doesn't fix it. I have no idea what's going on.

jpalardy:
hi @neddles 

Could you open an issue for this?

PR #187, opened 2019-02-08
Changes tmux syntax to avoid vim parsing error

defndaines:
With the previous syntax, using an array index, vim would report an error when editing your .vimrc:
```
E684: list index out of range: 0
```
Using the get syntax avoids this parsing confusion, although both work just fine.

Note, too, that this fixes a major headache I was having where vim-slime was working just fine from tmux (version 2.8), but didn't work inside tmate (version 2.2.1).


jpalardy:
Good catch, thanks 👍

PR #188, opened 2019-02-10
Default sessionname to $STY

RhysU:
GNU Screen sessionname defaults to the current session inside which Vim is running.  Permits just hitting enter on the default sessionname.

### DIFF
--- a/autoload/slime.vim
+++ b/autoload/slime.vim
@@ -34,7 +34,8 @@ endfunction
 
 function! s:ScreenConfig() abort
   if !exists("b:slime_config")
-    let b:slime_config = {"sessionname": "", "windowname": "0"}
+    " STY is the 'Alternative socket name' which is set inside GNU Screen
+    let b:slime_config = {"sessionname": "" . $STY, "windowname": "0"}
   end
   let b:slime_config["sessionname"] = input("screen session name: ", b:slime_config["sessionname"], "custom,<SNR>" . s:SID() . "_ScreenSessionNames")
   let b:slime_config["windowname"]  = input("screen window name: ",  b:slime_config["windowname"])


autoload/slime.vim
```diff
@@ -34,7 +34,8 @@ endfunction
 
 function! s:ScreenConfig() abort
   if !exists("b:slime_config")
-    let b:slime_config = {"sessionname": "", "windowname": "0"}
+    " STY is the 'Alternative socket name' which is set inside GNU Screen
+    let b:slime_config = {"sessionname": "" . $STY, "windowname": "0"}
```
RhysU: I am unclear if `"" . $STY` is preferred for handling a possibly unset $STY or if just $STY is sufficient and will return an empty string.  I tested both, but I'll admit that I didn't hammer on the tab completion aspect.  Using `"" . $STY` felt like it'd always give a string-valued result.  Suggestions much appreciated.

RhysU: I could also see this instead as a change to the "If you want vim-slime to prefill the prompt answers..." portion of the README.  Please shout if you'd prefer that.

jpalardy:
Hi there,

Isn't this a config change? Do we need to change `autoload.vim` for this?

I think this would work in your `.vimrc`:

    let g:slime_default_config  = {"sessionname": $STY, "windowname": "0"}

RhysU:
It could definitely be left as a config change if you prefer it that way.  In that case, please close this without merging and I'll add your suggestion above to the README in another merge request.

That said, using STY in this manner at least makes the default sessionname useful for some folks.  Presently its not useful for anyone (an empty string always requires editing vs STY where hitting enter is a well-formed input).

jpalardy:
The empty default works for completion (just press TAB).

Yes, let's make this a README pr. Thanks!

PR #189, opened 2019-02-14
Add the cell-sending functionality

AStcherbinine:
Just a few add to allow to use "Cell" like in Spyder or Pyzo e.g.
A cell can be delimited by `##`, `%%`, or `# %%`, and other delimiters can be setted up with
`let g:cell_delimiter`
You can send the cell to IPython using `<Ctrl-g>`

Inspired by _https://github.com/julienr/vim-cellmode_


jpalardy:
Hi @AStcherbinine ,

I'm not sure that belongs in vim-slime. Couldn't this be done with mappings?

- mapping to select a cell
- send with `c-c, c-c` normally

PR #195, opened 2019-03-31
Improve neovim terminal sliming.

a-vrma:
Use jobid 3 as default.
Use chansend function as jobsend is deprecated.

----

I chose jobid 3 as the default after some experimenting. It seems that it opens the first new terminal as 3. Obviously this is not perfect but it should improve the experience of people using neovim at least a little.


jpalardy:
Hi there,

This is looking good and I'm going to merge it.

Would you consider looking for more solid solution than using `3` for the jobid. (not sure what that would be, however)

Lattay:
Hello, I just made #196 and I went here two minutes later seing that my pull request kinda answer your concern @jpalardy. However I think there is still possible improvement (first of all using user configuration to avoid prompting the user each time).

PR #196, opened 2019-04-07
Provide the option of creating a new terminal with neovim

Lattay:
This commit ask the user if he wants to automatically create a new terminal in a vertical or horizontal split with a provided command. It extends the previous behavior where you had to set up the REPL beforehand and know the job id.
I did that to address my own usage and it is pretty simple implementation.
What do you think ?

### DIFF
--- a/autoload/slime.vim
+++ b/autoload/slime.vim
@@ -105,11 +105,36 @@ function! s:NeovimSend(config, text)
   call chansend(str2nr(a:config["jobid"]), split(a:text, "\n", 1))
 endfunction
 
+function! s:NeovimTermSplit(cmd, vertical) abort
+    if a:vertical
+        exec("vsplit term://" . a:cmd)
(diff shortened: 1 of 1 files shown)


jpalardy:
Hi @Lattay ,

Thanks for the PR.

Unfortunately, the behavior you propose is unlike that of other plugins (querying then opening the horizontal/vertical terminal). I believe the responsibility for doing those things falls outside vim-slime itself.

Lattay:
I understand, even if I think that since terminal is a quite recent feature, innovation is possible.
Is there a hook in the config process for the user (me) to put this kind of logic ?

jpalardy:
I think I would put this logic in the key binding. One mapping for horizontal, another for vertical … you could write a function to open the terminal, get its config and prefill the slime config. (you've already done most of the work in this PR)

PR #200, opened 2019-05-08
IPython bug fix

NedWestern:
- replace empty lines with a comment character
- send pause after pasting text
- send newline character after pause to execute


jpalardy:
Merging for now, thanks :-)

Keep an eye on issues coming in and let's see how it goes.

PR #202, opened 2019-05-08
Additional newline character to correctly enter cpaste mode

NedWestern:
New tact. Keep cpaste mode, but just add a newline so it works properly for IPython 7.2. Note, this is also required for pasting a single line.


NedWestern:
Damnit.. This was working, but sometimes it doesn't. It's totally a timing issue. Very frustrating. You need a way to control how IPython handles empty lines. Otherwise, replace empty lines with something else.

PR #205, opened 2019-07-22
Use tmux bracketed paste. Fixes #191.

### LINKED ISSUE #191: Paste comes with additional indentation with pgcli
Using vim-slime to send a paragraph like the following to [pgcli](https://github.com/dbcli/pgcli)

```sql
CREATE TABLE number_data_types (
    numeric_column numeric(20,5),
    real_column real,
    double_column double precision
);
```

It gets pasted, but for each new line there's additional indentation:

```
analysis> CREATE TABLE number_data_types (
     numeric_column numeric(20,5),
         real_column real,
             double_column double precision
             );
```

oblitum:
From `man tmux`:

> paste-buffer [-dpr] [-b buffer-name] [-s separator] [-t target-pane]
>               (alias: pasteb)
>         Insert the contents of a paste buffer into the specified pane.  If not specified, paste into the current one.  With -d, also delete the paste buffer.  When output, any linefeed (LF) characters in the paste buffer are re‐
>         placed with a separator, by default carriage return (CR).  A custom separator may be specified using the -s flag.  The -r flag means to do no replacement (equivalent to a separator of LF).  **If -p is specified, paste
>         bracket control codes are inserted around the buffer if the application has requested bracketed paste mode**.

### DIFF
--- a/autoload/slime.vim
+++ b/autoload/slime.vim
@@ -74,6 +74,9 @@ endfunction
 function! s:TmuxSend(config, text)
   call s:WritePasteFile(a:text)
   call s:TmuxCommand(a:config, "load-buffer " . g:slime_paste_file)
+  call s:TmuxCommand(a:config, "paste-buffer -p -d -t " . shellescape(a:config["target_pane"]))
+  call system("cat > " . g:slime_paste_file, "\n")
+  call s:TmuxCommand(a:config, "load-buffer " . g:slime_paste_file)
   call s:TmuxCommand(a:config, "paste-buffer -d -t " . shellescape(a:config["target_pane"]))
 endfunction
 


oblitum:
I just tried whether it affects `ipython` or `ptpython` and both are working fine. Can't say whether it's working better though, and ends up fixing any of the `ipython` issues.

jpalardy:
Looking good, thanks :+1

jpalardy:
I pulled locally, then reverted.

This introduces an extra newline that isn't needed in other modes...

jpalardy:
Scenario:

- open terminal, tmux, split vertical
- open vim, type some text -- `ls -l`
- slime over to other shell
- `ls -l` runs, plus extra newline

oblitum:
@jpalardy I'm not sure which line is this, can you apply a fix for that? #170 does it in a different way, which maybe doesn't add this newline.

PR #218, opened 2019-12-11
Add an example for `g:slime_vimterminal_config`

iblis17:


jpalardy:
Looking good, thank you :-)

Issue #220, opened 2019-12-18
Restore the window scroll position after SlimeParagraphSend


magv:
Hi, folks. Currently `SlimeParagraphSend` (i.e. c-c c-c) tries to save and restore the cursor position by calling `getcurpos()` and then `setpos()` (see [slime.vim:320](/jpalardy/vim-slime/blob/53c781b9532da50ab9d6ab1bbe640902b2629a0e/autoload/slime.vim#L325)). However if a part of the paragraph is not in the view of the current window, hitting c-c c-c will make the window scroll so that the whole paragraph is visible. I find this scrolling to be unexpected and mildly annoying. If this is not deliberate, then would it be possible to save/restore the scroll position too? Just using `winsaveview()` and `winrestview()` should be enough. Something like this maybe?

```diff
diff --git a/autoload/slime.vim b/autoload/slime.vim
index 58a0f9e..2de20fa 100644
--- a/autoload/slime.vim
+++ b/autoload/slime.vim
@@ -319,19 +319,13 @@ endfunction
 
 function! slime#store_curpos()
   if g:slime_preserve_curpos == 1
-    let has_getcurpos = exists("*getcurpos")
-    if has_getcurpos
-      " getcurpos() doesn't exist before 7.4.313.
-      let s:cur = getcurpos()
-    else
-      let s:cur = getpos('.')
-    endif
+    let s:cur = winsaveview()
   endif
 endfunction
 
 function! s:SlimeRestoreCurPos()
   if g:slime_preserve_curpos == 1 && exists("s:cur")
-    call setpos('.', s:cur)
+    call winrestview(s:cur)
     unlet s:cur
   endif
 endfunction
```

I think both functions are available starting at vim 7.0, so there's no need to check for their existence. Maybe these break some use case that I don't know about, but they seem to work fine for me.

jpalardy:
Hi @magv,

That sounds good, if you've tested it and you're happy with, send me a PR.

Thanks 😃

magv:
Thanks.

PR #221, opened 2019-12-19
Save and restore the full window scroll state

magv:
This solves issue #220.


jpalardy:
Tested locally, seems to work great.

Thanks for the patch.

PR #225, opened 2020-02-16
Add X11 support via xdotool

ynco:
Allow vim-slime to send text to any X11 window via the `xdotool` utility. `SlimeConfig` calls `xdotool selectwindow`, which lets the user identify a window by pointing at it with the mouse cursor. For a mouseless workflow, it's always possible to edit the configuration manually, so this seems a reasonable default.

### DIFF
--- a/README.md
+++ b/README.md
@@ -159,6 +159,16 @@ kitty target window
 
     This is the id of the kitty window that you wish to target. See e.g. the value of $KITTY_WINDOW_ID in the target window.
 
+### X11
+
+x11 is *not* the default, to use it you will have to add this line to your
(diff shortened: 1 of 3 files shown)


jpalardy:
Hi @ynco,

It looks good to me. My only concerns about this is discoverability and maintainability… I'm not sure at what point the documentation for different systems would be better served out of the main README… 🤔 

I'll merge this for now, but I might have to reach out in the future if I need help with reorganization.

Thanks

ynco:
> I'll merge this for now, but I might have to reach out in the future if I need help with reorganization.

Sure. Thanks!

PR #229, opened 2020-03-21
Added SlimeSendCell command

menganha:
New functionality to send blocks of code between two comment strings
(user specified).  This new features tries to emulate the cell-like mode of
popular notebook-type frontends (jupyter, ipython, etc.)

This commit adds the command SlimeSendCell and the function send_cell
which send code between the strings (usually a comment string) specified
in the global variable (by default unset) g:slime_cell_delimiter. 

The new functions/command are written in the same philosophy as the 
vim-slime plugin.

Things that can be improved/added:
* Possibly add a default cell delimiter for each file type (in ftplugin)
* Add this command and possible one example to the documentation

### DIFF
--- a/README.md
+++ b/README.md
@@ -282,6 +282,10 @@ By default, vim-slime will try to restore your cursor position after it runs. If
 
     let g:slime_preserve_curpos = 0
 
+If you want to send blocks of code between two delimiters, emulating the cell-like mode of REPL environments like ipython, matlab, etc., you can set the cell delimiter on the `g:slime_cell_delimiter` variable and use the `<Plug>SlimeSendCell` mapping to send the block of code. For example, if your are using ipython you could use the following:
+
+    let g:slime_cell_delimiter = "#%%"
+    nmap <leader>s <Plug>SlimeSendCell
 
 Language Support
 ----------------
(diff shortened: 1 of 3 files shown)


jpalardy:
Hi @menganha,

It's looking good :-)

But I think you're right:
- what would be a reasonable default for `g:slime_cell_delimiter`?
- what kind of documentation would be appropriate for this?

menganha:
Hi @jpalardy,

- I don't think it would make sense to have default value since this new functionality is intended for REPL style of programming. The delimiter is normally some kind of inline comment string, e.g., "#%%" for python or  "\\%%" for javascript, defines code "cells" that are sent in block to the interpreter.  I think that since vim-slime has a more general functionality the added `SlimeSendCell` mapping should not be part of the default "out of the box" config.
- I would just add to the documentation the additional mapping description with an additional note of how to configure it, i.e., to simply define the global variable `g:slime_cell_delimiter`

If you find this OK, I can work on a proposal for the documentation.

jpalardy:
Hi @menganha,

I think all that's missing is a blurb for the [Advanced Configuration](https://github.com/jpalardy/vim-slime#advanced-configuration) part of the README.

High-level:

- the goal: to grab (vertically) from one delimiter to the next
- example config (`g:slime_cell_delimiter` and key binding)
- sample where it would be used (code + delimiters)

And anything else you think would be useful.

menganha:
added blurb :)

jpalardy:
Looking good, thank you 👍 

If anything specific to this functionality comes up, I might call on you to help.

Thanks again.

PR #237, opened 2020-04-11
Allow override of *Config and *Send function with user-defined functions

jpalardy:
Update: function names changed ⚠️ 

Will be able to override a plugin's default Config and Send (either or both) functions:

```vim
" echo config
" call executable "send-somewhere" by piping text to its STDIN
function! SlimeOverrideSend(config, text)
  echom a:config
  call system("send-somewhere", a:text)
endfunction

" set "key" to (default) 1
function! SlimeOverrideConfig()
  let b:slime_config = {}
  let b:slime_config["key"] = input("key: ", 1)
endfunction

```

More documentation (examples) will be needed to flesh this out.

### DIFF
--- a/README.md
+++ b/README.md
@@ -21,13 +21,29 @@ Context for [SLIME](https://en.wikipedia.org/wiki/SLIME):
 
 Grab some text and send it to a [GNU Screen](http://www.gnu.org/software/screen/) / [tmux](https://tmux.github.io/) / [whimrepl](https://github.com/malyn/lein-whimrepl) / [ConEmu](http://conemu.github.io/) session / NeoVim Terminal / Vim Terminal
 
-    VIM ---(text)---> screen / tmux / whimrepl / ConEmu / NeoVim Terminal / Vim Terminal
+Grab some text and send it to a target, most probably [GNU Screen](http://www.gnu.org/software/screen/), [tmux](https://tmux.github.io/) or [Vim Terminal](https://vimhelp.org/terminal.txt.html).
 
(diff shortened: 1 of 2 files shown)


jpalardy:
@huijunchen9260,

I spent a lot of time thinking about how to fix your use case and other ones. I think this PR is a straightforward solution.

In short:
- the rest of vim-slime stays the same, no need to fork
- what most people wanted was a way to tweak logic, something close but not quite what their plugins already did
- being able to override with a custom function is VERY flexible

in your case, add this function to your `.vimrc`:

```vim
function! SlimeCustomConfig()
  if !exists("b:slime_config")
    let b:slime_config = {"window_id": ""}
  end
  let b:slime_config["window_id"] =system("name-of-external-script")
endfunction
```
or even: (if you don't use `g:slime_default_config`)

```vim
function! SlimeCustomConfig()
  let b:slime_config = {"window_id": ""}
  let b:slime_config["window_id"] = system("name-of-external-script")
endfunction
```

If we can make this work, I'll add some documentation and merge to master.

huijunchen9260:
Yeah it works. I add it into the `~/.vimrc`, and call this function by `:call SlimeCustomConfig()`, it indeed captured my script and do the job.

jpalardy:
@huijunchen9260,

I changed the function name prefixes from `SlimeOverride` to `SlimeCustom` ⚠️ 

I think that makes their purpose more obvious.

I'm working on the README, stay tuned.

huijunchen9260:
So now the cumtom function should be named override?

jpalardy:
yes, sorry for the change

PR #238, opened 2020-04-12
Dont send the delimiters while using the slime#send_cell function

mdedonno1337:
This small commit allows to send the current block without the delimiters with the send_cell fuction.

I think it make sense to remove the delimiters, since we usually use comments. The log in the target is cleaner this way.

What do you think?

### DIFF
--- a/autoload/slime.vim
+++ b/autoload/slime.vim
@@ -357,7 +357,7 @@ function! slime#send_cell(cell_delimiter) abort
   if !line_end
       let line_end = line("$")
   endif
-  call slime#send_range(line_ini, line_end)
+  call slime#send_range(line_ini+1, line_end-1)
 endfunction
 
 function! slime#store_curpos()
--- a/autoload/slime.vim
+++ b/autoload/slime.vim
@@ -357,7 +357,9 @@ function! slime#send_cell(cell_delimiter) abort
   " line before delimiter or bottom of file
   let line_end = line_end ? line_end - 1 : line("$")
 
-  call slime#send_range(line_ini, line_end)
+  if line_ini <= line_end
+    call slime#send_range(line_ini, line_end)
+  endif
 endfunction
 
 function! slime#store_curpos()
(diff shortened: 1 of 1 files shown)


autoload/slime.vim
```diff
@@ -357,7 +357,7 @@ function! slime#send_cell(cell_delimiter) abort
   if !line_end
       let line_end = line("$")
   endif
-  call slime#send_range(line_ini, line_end)
+  call slime#send_range(line_ini+1, line_end-1)
```
jpalardy, line 360: hi @mdedonno1337,

doing it unconditionally, it's off-by-one for the cases from line 354-359 — it would grab from line 2 and/or line (last - 1)

it would be more solid if the logic checked:
- if line_ini contains the cell_delimiter, +1
- if line_end contains the cell_delimiter, -1

alternatively, this could be done with else clauses for 354 and 357 🤔

mdedonno1337: mmmm good point, didn't thought about not using a delimiter for the top/bottom...
it's my first vim script, will try to do it!

thanks for the feedback!

mdedonno1337:
should do the trick (if I understood correctly the issue)

autoload/slime.vim
```diff
@@ -351,13 +351,17 @@ endfunction
...
   if !line_ini
       let line_ini = 1
+      let line_ini_delta = 0
   endif
   if !line_end
       let line_end = line("$")
+      let line_end_delta = 0
   endif
-  call slime#send_range(line_ini, line_end)
+  call slime#send_range(line_ini+line_ini_delta, line_end-line_end_delta)
```
jpalardy, line 364: how's this?

```vim
  " line after delimiter or top of file
  let line_ini = line_ini ? line_ini + 1 : 1
  " line before delimiter or bottom of file
  let line_end = line_end ? line_end - 1 : line("$")
  call slime#send_range(line_ini, line_end)
```

mdedonno1337: perfect !

jpalardy: tested locally, works for me … not sure I covered all the use-cases

- between 2 delimiters
- top of file and delimiter
- delimiter and bottom of file
- no delimiters (whole file)

@mdedonno1337 can you confirm?

mdedonno1337: Yes.

Added the check if the cell is empty (two separators, no content) or not.

jpalardy:
looking good, thank you 😄

PR #240, opened 2020-04-16
Terminate ipython %cpaste

trallarn:
This commit adds submission of "--\n" to terminate the ipython %cpaste command
so the code gets executed by the interpreter.


trallarn:
Closing since the error seems to be related to screen on Mac.

PR #241, opened 2020-05-04
Send separator when using the send_range function

mdedonno1337:
This PR allows to configure a variable (slime_sendrange_separator) in the vimrc to be send before the content of the range.

This allows to see easily the current chunk send to REPL.

Example:

image

### DIFF
--- a/autoload/slime.vim
+++ b/autoload/slime.vim
@@ -15,6 +15,10 @@ if !exists("g:slime_paste_file")
   let g:slime_paste_file = expand("$HOME/.slime_paste")
 end
 
+" separator send before sending a range
+if !exists("g:slime_sendrange_separator")
+  let g:slime_sendrange_separator = ""
+end
 """""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""
 " Screen
 """""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""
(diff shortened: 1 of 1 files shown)


jpalardy:
Hi @mdedonno1337,

Looks good! I checked the code and it does something different than what I expected… I thought it was going to include (or exclude) the delimiters (as in `SlimeSendCell`).

While I'm not against this feature, it's something that can be emulated outside vim-slime using a mapping and probably [SlimeSend1](https://github.com/jpalardy/vim-slime/blob/master/plugin/slime.vim#L12). In that case the user could have a binding that:
- always sends the delimiter and text
- (or) only sends the delimiter (without text)

What do you think?

mdedonno1337:
Well, still new to vimscript (second contribution as you know): did you meant something like that :
```
function! SlimeSendCellWithDelimiter() abort
  call slime#send(g:slime_sendrange_separator . "\r")
  call slime#send_cell(g:slime_cell_delimiter)
endfunction

nnoremap <Leader>z :call SlimeSendCellWithDelimiter()<CR>
```

in a `.vim` file in the plugin folder ?
It work's, yes.

I was wondering if you'd like to include this feature directly in slime (only if you'd like and think it is usefull ; it's your project).

mdedonno1337:
Closing the PR, OK for you ?

jpalardy:
Hi @mdedonno1337,

OK, let's close for now.

I think people can self-serve with `SlimeSend1` ~ though I wish we had better documentation for it…

PR #247, opened 2020-06-16
Add support for Jupyter QTConsole

adavidwilson:
Issue: Newlines were not copied to qtconsole.

Fix: Replace newlines with carriage returns if `g:slime_python_ipython_qtconsole` is set.

Very hasty testing indicates that it works, but it is by no means thorough. This is my first contribution ever, so let me know if there's anything I could improve.

### DIFF
--- a/ftplugin/python/README.md
+++ b/ftplugin/python/README.md
@@ -9,9 +9,18 @@ interpreter.
 [IPython](http://ipython.org/) has a `%cpaste` "magic function" that allows for
 error-free pasting. In order for vim-slime to make use of this feature for
 Python buffers, you need to set the corresponding variable in your .vimrc:
-
+```
     let g:slime_python_ipython = 1
-
+```
 Note: if you're using IPython 5, you _need_ to set `g:slime_python_ipython` for
-pasting to work correctly.
+pasting to work correctly. [Jupyter Console](https://jupyter-console.readthedocs.io/en/latest/)
+lacks %paste or %cpaste, so IPython should be used instead
+(see [jupyter/jupyter-console#20](https://github.com/jupyter/jupyter_console/issues/20)).
 
+[Jupyter QTConsole](https://qtconsole.readthedocs.io/en/stable/) has a quirk
+where newline characters are not read when pasting. In order for vim-silme to
+send carriage returns instead of newlines for QTConsole, set the following
+variable in your .vimrc:
+```
+    let g:slime_python_qtconsole = 1
+```
--- a/ftplugin/python/slime.vim
+++ b/ftplugin/python/slime.vim
@@ -14,10 +14,10 @@ function! _EscapeText_python(text)
     let except_pat = '\(elif\|else\|except\|finally\)\@!'
     let add_eol_pat = '\n\s[^\n]\+\n\zs\ze\('.except_pat.'\S\|$\)'
     let eol_lines = substitute(dedented_lines, add_eol_pat, "\n", "g")
-    if exists('g:slime_python_ipython_qtconsole')
-        return substitute(eol_lines, '\n', '\r', "g")
-    else
-        return eol_lines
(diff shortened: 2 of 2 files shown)


jpalardy:
Hi @adavidwilson,

Thanks for the contribution 😄 

I'm wondering what's different about the qtconsole… (or python, in this case). Could you confirm what your vim-slime configuration looks like:
- which vim version?
- which target? (tmux, screen …) which version?
- anything else that's special about your config?

adavidwilson:
Hi @jpalardy,

Thanks for the words of support and the great plugin!

The qtconsole is a separate interactive console implemented in PyQt5 and pops up as its own X11 window. I was aware of previous issues specific to the ipython interpreter and tried a bunch of the recommendations in those, but this issue seemed different. Whenever sliming a region, it would just concatenate all the lines together. Running ipython in another terminal window worked fine, but lacks the interactivity I'm looking for (inline plots).

The only slime-related configs in my init.vim are `let g:slime_target = "x11"` and `let g:slime_python_ipython_qtconsole = 1`. As for my setup, I hope this covers everything:

- Vim Version: Neovim v0.4.3
- Plugin Manager: Vim Plug
- Slime Target: x11 (version 1.20.8-2)

Using Neovim terminal and x11 as the targets to other REPLs worked without a hitch. The problem was isolated to qtconsole.

jpalardy:
Hi @adavidwilson,

Does it work with other targets? (tmux, in particular)

adavidwilson:
Hi @jpalardy 

Using tmux as the target works with the regular python and ipython interpreters, but qtconsole is its own X window. This might be a newbie comment (I have yet to become comfortable in tmux), but I'm unaware of whether or not tmux can run X in a pane, so I can't test whether or not sliming to the qtconsole works when using tmux as the target.

Here are some screenshots of the issue:

- Before
before

- After
after

jpalardy:
hi @adavidwilson,

I hadn't realized that QTConsole wasn't running in a terminal (that's how things were _before_ the x11 target)

I'm very hesitant to change the python plugin (which a lot of people use, but I don't), especially for this specific combination {x11, qtconsole}

Can you have a look at [overrides](https://github.com/jpalardy/vim-slime#advanced-configuration-overrides) and see if can make it work? You could create `SlimeOverride_EscapeText_python` and replace in there.

adavidwilson:
Hi @jpalardy,

Is there any way I could make it more suitable for merging? If not, I completely understand. I can make it work for myself by just copy pasting my changes into an override.

My main motivation for submitting a PR was so others might benefit. Specifically, QtConsole + vim-slime offers a very enticing alternative to Jupyter Notebooks or Jupyter Lab that doesn't require a browser and supports vim bindings natively for an interactive Python workflow.

jpalardy:
Hi @adavidwilson,

I'm having a similar discussion [here](#248).

The python plugin is the most complicated and nothing getting simpler; it has been filled with special cases and I'm just trying to see how we can extricate ourselves from this situation without having to provide a flowchart to help users decide which options they need…

I can think of many options:
- maybe we need a post-language-plugin filter — there's nothing python-specific about your qtconsole fix (it's just newlines and carriage returns)
- maybe this needs a fix upstream, in the qtconsole repo — arguably not a quick fix
- maybe there's enough python logic (and configurations) to justify a command-line "fixer" (which supports (and tests!) all permutations)
- maybe we need documented "recipes" (a wiki?) for specific configurations

I'm open to suggestions; especially if they don't involve "one more if" and "one more config" (+ documentation update)

cc: @harritaylor

adavidwilson:
Hi @jpalardy,

Understood. I'll close my PR for now then and try to work on coming up with a less intrusive fix or a wiki page documenting the issue. It took quite a bit of experimentation to realize that would fix it. Thank you for the constructive comments :relaxed:

al-jshen:
@adavidwilson as someone switching over to qtconsole from notebooks, I just wanted to thank you for your solution :)

PR #248, opened 2020-06-17
More support for iPython & Jupyter console REPLs

harritaylor:
So there may be a better way of doing this, but I avoided using variables so that the code is as explicit as possible. This PR covers two things: 

* Support for [vi editing mode](https://ipython.readthedocs.io/en/stable/config/options/terminal.html?highlight=vi%20#configtrait-TerminalInteractiveShell.editing_mode) by sending `escape, i` before any text is pasted.
* Trimmed `--` that is appended to each paste for `jupyter console`, whilst still utilising `%cpaste`.

Unfortunately, it does include two extra config options, `g:slime_python_jupyter` and `g:slime_python_vi_repl`. I haven't documented them in the readme, as I wanted to hear your feedback regarding the PR first. 

Cheers!

### DIFF
--- a/ftplugin/python/slime.vim
+++ b/ftplugin/python/slime.vim
@@ -5,7 +5,15 @@ end
 
 function! _EscapeText_python(text)
   if exists('g:slime_python_ipython') && len(split(a:text,"\n")) > 1
-    return ["\e i", "%cpaste -q\n", g:slime_dispatch_ipython_pause, a:text, "--\n"]
+    if exists('g:slime_python_vi_repl')
+        return ["\e i","%cpaste -q\n", g:slime_dispatch_ipython_pause, a:text, "--\n"]
+    else
+        return ["%cpaste -q\n", g:slime_dispatch_ipython_pause, a:text, "--\n"]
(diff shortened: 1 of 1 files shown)


jpalardy:
Hi @harritaylor,

It looks like that would work … but the python plugin is already highly customized. (and [more](#247) possible)

Can you take a look at [overrides](https://github.com/jpalardy/vim-slime#advanced-configuration-overrides)?

If the number of permutations necessary to make python work keeps its current trend, maybe it's worth thinking about breaking off that plugin to its own project? 🤔

harritaylor:
Hey, yeah I agree, I don't like the amount of extra variables needed for it... I thought it would be a quick small fix but it's actually quite convoluted! 

> If the number of permutations necessary to make python work keeps its current trend, maybe it's worth thinking about breaking off that plugin to its own project? 🤔
 
Perhaps! There's all sorts of edge cases it seems. I'll see how I get on with overrides for my personal uses and revisit in the future if it seems worth it. Thanks!

jpalardy:
Ok, let me know 👍

Issue #208, opened 2019-07-27
Commands not executing in IPython with SlimeSend1 and SlimeSendCurrentLine


hanschen:
`SlimeSend1` and `SlimeSendCurrentLine` used to send the specified text/current line to IPython and then execute the line. Now they send the line and then append a carriage return without executing the line.

This issue seems to have been introduced by #206.

**To reproduce**

1. Open tmux, open vim in a tmux pane and IPython in another tmux pane.
2. Run `:SlimeSend1 1` in vim.

**Expected behavior**

    In[1]: 1
    Out[1]: 1

The line (`1`) is executed in IPython.

**Actual behavior**

    In [1]: 1
        ...:

A newline is appended to the line instead of executing the line.

**Other information**

Reproduced with neovim 0.4.0 and vim 8.1 using tmux 2.9a and newest version of vim-slime (commit 705990815e7874cabf13c77a41462f1326d04ca2).

vim-slime is configured as follows:

    let g:slime_target = 'tmux'

jpalardy:
Hi @hanschen,

I [reverted](https://github.com/jpalardy/vim-slime/commit/dc8ca22cef3f87999d926e18e2e230145e013838) to before #206. Can you confirm that it fixes this issue?

@oblitum ^

oblitum:
:-/

Sad this feature is ended up being reverted. I guess the closed issues related will at least need to get reopened, as it is slime doesn't work with `pgcli`.

hanschen:
@jpalardy Yes, that fixes this issue. The latest vim-slime (dc8ca22cef3f87999d926e18e2e230145e013838) works as expected.

jpalardy:
@hanschen ok, I'm going to close this for now.

@oblitum I'm open to suggestions on #191. In the meantime, how do you feel about putting the reverted code on a branch?

oblitum:
> @oblitum I'm open to suggestions on #191. In the meantime, how do you feel about putting the reverted code on a branch?

I'm currently using `Plug 'jpalardy/vim-slime', {'commit': '7059908'}`.

oblitum:
> @oblitum I'm open to suggestions on #191.

Let's reopen it  as first step :-/

oblitum:
hey @jpalardy, I finally got bit fed up with being stuck at 7059908 and gave a look at this issue, you can't imagine how easy it's the fix, it came as soon as I opened slime source and notice this:

```vim
[...]

Issue #245, opened 2020-05-26
[vim-slime + tmux + vim + ipython] Issue when copy multiple lines; newline character appears as ^M


pvmilk:
I have issue when using vim-slime to send python code to ipython.
It only happens when I send multiple line of codes (more than 1 line). 
The newline character simply appear as ^M in the ipython screen. 
It seems that the format of code was correct, just that the newline is converted to DOS/Windows newline character. 

![image](https://user-images.githubusercontent.com/1500082/82925432-ab012800-9fb8-11ea-80d1-e043e768ba72.png)

In addition
- It doesn't not always happen. Usually, it works fine initially. After sending some multiple line of codes for several times, it appears. 
- Also, my ipython interpreter is in the docker environment. Usually, after I restart the docker container, it works fine again. 

I knew that I had a quite complicated setup, so I am trying to debug by myself also. 
I think newline character was converted to ^M at some buffer, but I don't which one. 
Can you help suggesting where should I look?

ENV:
- [host] vim-slime commit a81139239d6447b1c021e4873e6a14b47f32b65a
- [host] VIM - Vi IMproved 8.2 (2019 Dec 12, compiled May  2 2020 12:57:56)
- [host] tmux 2.7
- [docker] Python 3.6.9 (default, Apr 18 2020, 01:56:04)
- [docker] IPython 7.13.0

jpalardy:
Hi @pvmilk,

There's surprisingly little relevant code in vim-slime 😄 

- [tmux send code](https://github.com/jpalardy/vim-slime/blob/master/autoload/slime.vim#L64)
- [python ftplugin code](https://github.com/jpalardy/vim-slime/blob/master/ftplugin/python/slime.vim)

As for the best way to debug, a generous use of `:echom` and `:messages` ([details](https://learnvimscriptthehardway.stevelosh.com/chapters/01.html))

Let me know what you find and if I can help.

jessebett:
I'm also having this issue. I have also set `:let g:slime_python_ipython=1`.

pvmilk:
@jessebett Finally after a long search, I found that you can use [Ctrl + d] to exit that `%cpaste` session. You might need to press it a couple of time and wait a few seconds, but it definitely better than the need to kill the process (which I also used that for a while also). 

I am still searching for where the conversion occurred. Probably will take a while as I am not so familiar with vimscript and tmux buffer. Will report back.

jessebett:
[...]

PR #254, opened 2020-07-01
Use tmux bracketed paste

### LINKED ISSUE #191: Paste comes with additional indentation with pgcli

oblitum:
Fixes #191, #208.

Originally from #170.

Dunno what happened when discussing at #208, but this is actually working on
top of master. Maybe I got issues because of the state of the `.slime_paste`
file while I was trying changes.

Please give it a try :)

### DIFF
--- a/autoload/slime.vim
+++ b/autoload/slime.vim
@@ -72,9 +72,19 @@ function! s:TmuxCommand(config, args)
 endfunction
 
 function! s:TmuxSend(config, text)
-  call s:WritePasteFile(a:text)
+  let l:lastchar = a:text[len(a:text)-1]
+  if l:lastchar == "\n"
+    call s:WritePasteFile(a:text[0:(len(a:text)-2)])
+  else
+    call s:WritePasteFile(a:text)
+  endif
+  call s:TmuxCommand(a:config, "send-keys -X -t " . shellescape(a:config["target_pane"]) . " cancel")
   call s:TmuxCommand(a:config, "load-buffer " . g:slime_paste_file)
-  call s:TmuxCommand(a:config, "paste-buffer -d -t " . shellescape(a:config["target_pane"]))
+  call s:TmuxCommand(a:config, "paste-buffer -p -d -t " . shellescape(a:config["target_pane"]))
+  if l:lastchar == "\n"
+    call s:TmuxCommand(a:config, "set-buffer \<CR>")
+    call s:TmuxCommand(a:config, "paste-buffer -d -t " . shellescape(a:config["target_pane"]))
+  endif
 endfunction
 
 function! s:TmuxPaneNames(A,L,P)
--- a/plugin/slime.vim
+++ b/plugin/slime.vim
@@ -9,9 +9,9 @@ let g:loaded_slime = 1
 
 command -bar -nargs=0 SlimeConfig call slime#config()
 command -range -bar -nargs=0 SlimeSend call slime#send_range(<line1>, <line2>)
-command -nargs=+ SlimeSend1 call slime#send(<q-args> . "\r")
+command -nargs=+ SlimeSend1 call slime#send(<q-args> . "\n")
 command -nargs=+ SlimeSend0 call slime#send(<args>)
-command! SlimeSendCurrentLine call slime#send(getline(".") . "\r")
+command! SlimeSendCurrentLine call slime#send(getline(".") . "\n")
 
 noremap <SID>Operator :<c-u>call slime#store_curpos()<cr>:set opfunc=slime#send_op<cr>g@
 


autoload/slime.vim
```diff
@@ -72,9 +72,19 @@ function! s:TmuxCommand(config, args)
...
+    call s:WritePasteFile(a:text)
+  endif
+  call s:TmuxCommand(a:config, "send-keys -X -t " . shellescape(a:config["target_pane"]) . " cancel")
   call s:TmuxCommand(a:config, "load-buffer " . g:slime_paste_file)
-  call s:TmuxCommand(a:config, "paste-buffer -d -t " . shellescape(a:config["target_pane"]))
+  call s:TmuxCommand(a:config, "paste-buffer -p -d -t " . shellescape(a:config["target_pane"]))
+  if l:lastchar == "\n"
+    call s:TmuxCommand(a:config, "set-buffer \<CR>")
+    call s:TmuxCommand(a:config, "paste-buffer -d -t " . shellescape(a:config["target_pane"]))
+  endif
```
jpalardy, line 87: trying to follow here:
- if lastchar is a newline, send everything before it…
- then send `cancel` ?
- if lastchar is a newline, send a carriage-return?

plugin/slime.vim
```diff
@@ -9,9 +9,9 @@ let g:loaded_slime = 1
 
 command -bar -nargs=0 SlimeConfig call slime#config()
 command -range -bar -nargs=0 SlimeSend call slime#send_range(<line1>, <line2>)
-command -nargs=+ SlimeSend1 call slime#send(<q-args> . "\r")
+command -nargs=+ SlimeSend1 call slime#send(<q-args> . "\n")
 command -nargs=+ SlimeSend0 call slime#send(<args>)
-command! SlimeSendCurrentLine call slime#send(getline(".") . "\r")
+command! SlimeSendCurrentLine call slime#send(getline(".") . "\n")
```
jpalardy, line 14: this would have consequences beyond the tmux plugin…

autoload/slime.vim
oblitum: To be honest I just recovered the old pull here, my knowledge is equal to or less than yours.

plugin/slime.vim
oblitum: If you have a suggestion to tailor it just for tmux I can improve the pull, I only ever use vim-slime with tmux and nothing else.

jpalardy: hi @oblitum,

This is tricky. I try to keep an eye on what problems people report and how often these people happen. Then, I need to keep in mind "what could break" for any given change … and how many people would be affected. (it doesn't help that there's an explosion of configurations out there: which vim, which OS, which terminal, which REPL, etc.)

I can leave this PR open and see how many people it helps (or not).

In the meantime, I tried to fix a workaround for customizations without having people maintain their own fork: [overrides](https://github.com/jpalardy/vim-slime#advanced-configuration-overrides).

oblitum: Okay thanks. Actually this pull does the job for me already, can't be easier than what's already written right? I then can stick to my fork and occasionally rebase this commit on top of upstream... I don't have much knowledge on vim-slime code base, so I thought that adding back something that was on master before wouldn't be big deal.

oblitum: On a side note, there's some issues involving tmux open, given that bracketed paste changes how tmux is handled substantially, _maybe_ it can make vim-slime better (or worse) regarding those issues.

oblitum: Like #245, but it would require that reporters there give this a try.

carlocab:
It seems this PR fixes an issue I have when using vim-slime with Julia: KristofferC/OhMyREPL.jl#224. (I've only tried the thing that doesn't work on master, but I don't mind testing it a bit more to see if it's broken something else.)

May I ask what the status of this PR is? I see you're currently intending for users to use [overrides](https://github.com/jpalardy/vim-slime#advanced-configuration-overrides) to customise the behaviour they need, but I'm not quite sure how to do that in this case.

I suppose I could figure it out, if you're not intending to merge this. All I really need is for slime to use bracketed-paste when `ft=julia`.

oblitum:
@carlocab from my side I don't expect to go much further beyond what's up, currently I'm happy enough with my fork, but would be great if tmux bracketed paste would eventually be supported upstream, I suppose it fixes lots of problems with many unknown REPL clients, like it does for pgcli and, as you say, julia.

carlocab:
@oblitum that's good to know, thanks. I'm happy with what you've done with it, and I'll probably stick with it while your changes haven't been merged. However, I will mention that your PR currently breaks `<Plug>SlimeSendCell`. Doing that with [your tmux-bracketed-paste branch](https://github.com/oblitum/vim-slime/tree/tmux-bracketed-paste) gives me `E119: Not enough arguments for function: slime#send_cell`.

That said, I'd rather have bracketed-paste even if it means I can no longer use `<Plug>SlimeSendCell`. Happy to help figure out what's causing the error if needed, however.

jpalardy:
Hi @oblitum, @carlocab,

I left this open because I didn't know how to integrate it. Here's a proposal:
- let's make a new mode "tmux-bracketed" (maybe?)
- let's put your code there (and leave the current `tmuxSend` as is)
- let's add a note to the README about this mode, and when to use it

That also assumes I can lean on you if problems appear related to this.

How does that sound?

oblitum:
@jpalardy to be honest I'm not sure about that, I think you have indeed pointed right regarding the consequences for the `\r`/`\n` change. I don't know the codebase very well and don't plan to dig it much, so, I myself don't know about "the consequences", since I just use it with tmux. Ideally I wished there was a patch that incorporated, this but that played well with the rest of the code, without such side effects. In my view, incorporating it like you propose doesn't feel pretty, such mode then would only guaranteed to work for tmux and broken for then rest? So I guess that should be stated upfront.

oblitum:
You may count on me on minor fixes or revert regarding such change, I just can't fix it for non-tmux case.

jpalardy:
fair enough 👍 

if you think it's helpful, I would accept a PR to point to your fork with a relevant explanation.

carlocab:
@jpalardy your proposal seems ideal, though I guess I'd want to see what the details look like before I know for sure.

You can count on me for assistance in maintaining this part of the code, including problems related to it. However, I will mention that I currently know very little Vimscript, so I may not be able to provide much help. I do intend on learning a bit more of it though, so hopefully, I'm able to help out more over time.

As an aside, would the issue with `<Plug>SlimeSendCell` be related to the `\r`/`\n` change?

jpalardy:
Hi @carlocab 

> would the issue with <Plug>SlimeSendCell be related to the \r/\n change?

it certainly sounds like it… you can try changing back-and-forth locally to see what works.

Let me know

carlocab:
Ok, for some reason I can't reproduce the error I reported earlier. However, I can say that `<Plug>SlimeSendCell` does not work in @oblitum's fork -- nothing happens when I try to call it. This is the case even if I change the `\n` back to `\r`.

I deactivated all my plugins on `nvim` (v. 0.4.4) when I did this, so I would guess it's not something else that's causing the problem.

carlocab:
I experimented with this line:

https://github.com/jpalardy/vim-slime/blob/db486eaa39f14d130ddf6338aaa02127aa04b272/autoload/slime.vim#L77

If I change it to
```
call s:TmuxCommand(a:config, "paste-buffer -d -p -t " . shellescape(a:config["target_pane"]))
```
then it works perfectly for me. `<Plug>SlimeSendCell` also works as expected.

Of course, this may have an impact on your other users, so I guess the `-p` option needs to be opted into in a user's configuration. I guess it should be picked up somehow by `TmuxConfig`, perhaps using the value of `b:tmux_bracketed`. Then the paste command issued by `TmuxSend` would be `"paste-buffer -d -t "` or `"paste-buffer -d -p -t "` as determined by `TmuxConfig`.

How does that sound? If it works for you, I could try to carve out some time this weekend to implement it. (I know basically no vimscript, so you may have to bear with me a little.)

jpalardy:
I'm hesitant to recommend anything one way or the other… 😐 

```
     paste-buffer [-dpr] [-b buffer-name] [-s separator] [-t target-pane]
           (alias: pasteb)
         Insert the contents of a paste buffer into the specified pane.
         If not specified, paste into the current one.  With -d, also
         delete the paste buffer.  When output, any linefeed (LF) charac-
         ters in the paste buffer are replaced with a separator, by
         default carriage return (CR).  A custom separator may be speci-
         fied using the -s flag.  The -r flag means to do no replacement
         (equivalent to a separator of LF).  If -p is specified, paste
         bracket control codes are inserted around the buffer if the
         application has requested bracketed paste mode.
```

so `-p` looks like it wouldn't hurt (if it were always on), but it's hard to tell. I can give this a try locally for a few days and see if I notice anything different.

My concerns were about:
- the send-keys cancel — why is that needed?
- managing the \r and \n … 😫

jpalardy:
You mentioned julia … what other REPLs are representative?

(sometimes I wish vim-slime had good functional tests…)

carlocab:
I don't know much about other REPLs, but my guess is that bracketed paste would've fixed a bunch of issues you've had with Python/IPython too. I don't really understand the send-keys cancel and the \r and \n thing either.

I've read that bit of the manual too, though it is unfortunately not very explicit about what `if the application has requested bracketed paste mode` means. (And if tmux can tell, why does it need its user to specify the `-p` flag?) However, it does seem like some people would like the option of turning bracketed-paste off: [[1]](https://askubuntu.com/questions/662222/why-bracketed-paste-mode-is-enabled-sporadically-in-my-terminal-screen) [[2]](https://stackoverflow.com/questions/42212099/how-do-i-disable-the-weird-characters-from-bracketed-paste-mode-on-the-mac-os) [[3]](https://stackoverflow.com/questions/41832287/how-to-disable-wrapping-pasted-text-in-search-line) [[4]](https://stackoverflow.com/questions/33452870/tmux-bracketed-paste-mode-issue-at-command-prompt-in-zsh-shell). That said, I think it is generally quite useful [[5]](https://cirw.in/blog/bracketed-paste). It's probably better to give users the option to turn it on or off.

Alternatively, I guess you could just show an example of an override that uses bracketed-paste and let users use it if they wished. I haven't managed to figure out how to set up the override only for specific filetypes though. (I imagine I'll get there at some point.)

jpalardy:
I was reading number 5, I think this is it:

Screen Shot 2020-09-17 at 20 34 19

so tmux knows you're in bracketed mode because it saw a `\e[?2004h` at some point (toggling it on)

the `-p` flag to `paste-buffer` explicitly tells tmux to wrap the buffer with `\e[200~` and `\e[201~`

I would guess that `-p` is a no-op otherwise and would be OK to always be turned on.

carlocab:
Ok, sounds good to me.

I'll try to find an application that doesn't support bracketed paste and see what `tmux` does with it when the `-p` flag is active. That might help clarify the impact of leaving it always on.

oblitum:
Hi folks. Sorry, kinda busy lately.

>  I don't really understand the send-keys cancel and the \r and \n thing either.

@carlocab This PR was resurrected after a long time, to add that `\r`/`\n` thing as an attempted solution to the issue (#208) that caused the revert to the "original" PR. If you want to dig into the history of this, just check the threads I refer to in this PR.

oblitum:
I'm closing this PR as it doesn't play well with non-tmux usecases, and I don't know how to conciliate that for lack on understanding of the codebase.

For anyone interested to pick bracketed paste mode for tmux, feel free to pick from my fork until this gets covered upstream.

Issue #252, opened 2020-06-27
Support for Scala Ammonite REPL?


LeoKavanagh:
Hi,

I'm using Vim-Slime (normal Vim, with tmux) with Scala's [Ammonite REPL](http://ammonite.io/), an IPython-style improved Scala REPL.

The base Scala REPL you get from `scala` or `sbt console` treats a `^D` that follows `:paste` as "exit paste mode", but Ammonite only ever treats it as "exit the REPL entirely". (As a side note, Ammonite doesn't have a `:paste` command (see [this issue](https://github.com/lihaoyi/Ammonite/issues/81)), but rather wraps multi-line expressions in curly braces.)

The [Slime function for Scala](https://github.com/jpalardy/vim-slime/blob/master/ftplugin/scala/slime.vim) is set up to work with the base console, and therefore doesn't work well with Ammonite.

I was wondering if an IPython-style flag and would be welcome here too. That is, get the user to add the following line to their `.vimrc`:

```
let g:slime_scala_ammonite = 1
```

and also add an if-statement to the `_EscapeText_scala` function to check for it. Something like:
```
function! _EscapeText_scala(text)
   if exists('g:slime_scala_ammonite')
     return ["{\n", a:text, "}\n"]
   else
     return [":paste\n", a:text, "^D"]
   end
endfunction   
```

I've been using this version of the function personally for the past few days, and it seems to be working fine. Unfortunately I don't know enough VimScript to offer anything like a passing unit test.

Curious to get your thoughts on this. I'm happy to submit a PR if you think it would be a useful addition.

jpalardy:
Hi @LeoKavanagh,

If it works, and you're willing to help out with (potential) incoming issues, then I would be willing to merge that in 😀 

This would probably need a documentation patch … at least to document the new flag.

LeoKavanagh:
Sounds good! I'll get to work on a PR.

vando026:
Hi both, I followed this thread with interests, I have one issue and a suggestion. 

Suggestion: after checking its existence, shouldn't the _EscapeText function also include an `if g:slime_scala_ammonite==1` condition, as in 

```
function! _EscapeText_scala(text)
  if exists('g:slime_scala_ammonite')
    if g:slime_scala_ammonite==1
      return ["{\n", a:text, "}\n"]
    end
  end
  return [":paste\n", a:text, "C-d"]
endfunction

```
which does the first return statement if the variable exists and is equal to 1. 

[...]

PR #255, opened 2020-07-02
Add Ammonite support to Scala ftplugin

LeoKavanagh:
Addresses #252 

Adds support for Scala's Ammonite REPL if the user sets `let g:slime_scala_ammonite = 1` in their `.vimrc`.

Includes references to the options in the main README, and adds a new README explaining the difference in `ftplugin/scala`.

### DIFF
--- a/ftplugin/scala/slime.vim
+++ b/ftplugin/scala/slime.vim
@@ -1,5 +1,9 @@
 
 function! _EscapeText_scala(text)
-  return [":paste\n", a:text, ""]
+  if exists('g:slime_scala_ammonite')
+    return ["{\n", a:text, "}\n"]
+  else
+    return [":paste\n", a:text, ""]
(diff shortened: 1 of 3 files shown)


ftplugin/scala/slime.vim
```diff
@@ -1,5 +1,9 @@
 
 function! _EscapeText_scala(text)
-  return [":paste\n", a:text, ""]
+  if exists('g:slime_scala_ammonite')
+    return ["{\n", a:text, "}\n"]
+  else
+    return [":paste\n", a:text, ""]
+  end
```
jpalardy, line 7:
```suggestion
  end
  return [":paste\n", a:text, ""]
```

if returns, no else?

jpalardy:
Thanks for your help 👍

Issue #256, opened 2020-08-03
<Plug>SlimeSendCell breaks if g:slime_cell_delimeter is defined using an autocommand


mroavi:
The `<Plug>SlimeSendCell` mapping added in 6755b54b66f565dcf6bb5c600af04aeb345af531 doesn't seem to be working at the moment. The other mappings like `<Plug>SlimeRegionSend` and `<Plug>SlimeLineSend` do work fine.

Version: NVIM v0.5.0-577-g48ac77a14
vim-slime sha: f7442cdb1fb37778ddff93649e172b0ad752c237
OS: Ubuntu 18.04.4 LTS
Target: tmux 2.6

jpalardy:
Hi @mroavi,

In my [config file](https://github.com/jpalardy/dotfiles/blob/master/vim/plugins.vim#L30), I've been using this:

```vim
let g:slime_cell_delimiter = "#--"
nmap <c-c>d <Plug>SlimeSendCell
```

and it works.

What's your config? And what behavior do you see?

mroavi:
This is my config:
```vim
let g:slime_target = "tmux"
let g:slime_default_config = {"socket_name": "default", "target_pane": "{right-of}"}
let g:slime_dont_ask_default = 1
let g:slime_no_mappings = 1
let g:slime_cell_delimiter = "%%"
nmap <Leader>cc <Plug>SlimeSendCell
nmap <Leader>ls <Plug>SlimeLineSend
```
I tested with a different delimeter but it still dosen't work. I'm testing this on matlab/octave files.

The behavior I see is that after I press the `<Leader>cc` nothing occurs, i.e. no text is sent to tmux. On the other hand, if I type `<Leader>ls` then the current line is sent. I replaced `<Plug>SlimeSendCell` with `:echo "Test"<CR>` to make sure the key binding is working properly.

mroavi:
I found the problem: I had the `let g:slime_cell_delimiter` definitions inside several *filetype plugin* files (e.g. *ftplugin/octave.vim*). Apparently these files are sourced after the *plugin/slime.vim* file is sourced. This meant that the content of the following `if` block was not being executed:
```vim
if exists("g:slime_cell_delimiter")
  noremap <unique> <script> <silent> <Plug>SlimeSendCell :<c-u>call slime#send_cell(g:slime_cell_delimiter)<cr>
endif
``` 
[...]

PR #257, opened 2020-08-05
Allow definition of g:slime_cell_delimeter using autocommands

mroavi:
Fixes https://github.com/jpalardy/vim-slime/issues/256

### DIFF
--- a/autoload/slime.vim
+++ b/autoload/slime.vim
@@ -15,6 +15,10 @@ if !exists("g:slime_paste_file")
   let g:slime_paste_file = expand("$HOME/.slime_paste")
 end
 
+if !exists("g:slime_cell_delimiter")
+  let g:slime_cell_delimiter = ""
+end
+
 """""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""
 " Screen
 """""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""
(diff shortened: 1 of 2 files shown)


mroavi:
I initialize `g:slime_cell_delimeter` with an empty string. If `slime#send_cell` is called without the user assigning a delimiter string other than an empty string, a custom error will be echoed. Otherwise, the function will run as before.

autoload/slime.vim
```diff
@@ -349,16 +353,20 @@ function! slime#send_lines(count) abort
...
+    " line before delimiter or bottom of file
+    let line_end = line_end ? line_end - 1 : line("$")
 
-  if line_ini <= line_end
-    call slime#send_range(line_ini, line_end)
+    if line_ini <= line_end
+      call slime#send_range(line_ini, line_end)
+    endif
+  else
+    echoerr "g:slime_cell_delimiter is empty"
```
jpalardy, line 369: would it make sense to "prompt" the user here?

mroavi:
> if you're willing, I can open a PR against your branch and we can iterate on this
> 
> or, I can start a branch on my side and pull your changes

Sure. Let's go for the second option.

autoload/slime.vim
mroavi: I would rather not. A trend that I've seen in many vim plugins is that the state of the program should stay in sync with the state defined in the configuration files.

jpalardy:
have a look at #258

PR #258, opened 2020-08-06
Improve cell delimiter behavior

jpalardy:
cc: @mroavi

I started from your PR, but I think I simplified the general flow:

- no need to declare (even empty-default) a `g:slime_cell_delimiter`
- grab first of `b:slime_cell_delimiter` or `g:slime_cell_delimiter`, per ftplugin overrides
- early error if no `slime_cell_delimiter` available

I think all that's missing is some documentation clarification. Thoughts?

### DIFF
--- a/README.md
+++ b/README.md
@@ -294,6 +294,8 @@ If you want to send blocks of code between two delimiters, emulating the cell-li
     let g:slime_cell_delimiter = "#%%"
     nmap <leader>s <Plug>SlimeSendCell
 
+⚠️  it's recommended to use `b:slime_cell_delimiter` and set the variable in `ftplugin` for each relevant filetype.
+
 
 Advanced Configuration: Overrides
 ---------------------------------
--- a/autoload/slime.vim
+++ b/autoload/slime.vim
@@ -15,10 +15,6 @@ if !exists("g:slime_paste_file")
   let g:slime_paste_file = expand("$HOME/.slime_paste")
 end
 
-if !exists("g:slime_cell_delimiter")
-  let g:slime_cell_delimiter = ""
-end
-
 """""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""
 " Screen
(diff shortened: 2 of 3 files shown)


mroavi:
I think these changes are good: Getting rid of the global variable and explicitly denoting that the local `b:slime_cell_delimeter` has priority over the global one. What I would also change is the error message to:
```vim
    echoerr "Neither b:slime_cell_delimeter nor g:slime_cell_delimiter are defined"
```

mroavi:
or
```vim
echoerr "slime_cell_delimeter is not defined"
```
which is less verbose.

mroavi:
One thing to consider is if we should completely get rid of the global variable `g:slime_cell_delimeter`. If you take a look at tpope's *vim-commentary* package, he does not have a global counterpart for filetype specific options (e.g. `b:commentary_format`).

jpalardy:
In retrospect, `g:slime_cell_delimiter` is a problem — it doesn't seem to make sense for the _specific_ use case of cell delimiters.

Maybe there's a marginal case for "I only use cell_delimiter for filetype X, so global is fine"

However, I'm having trouble imagining how to deprecate it without breaking existing installations. 🤔 

Unless you have a workaround, I'll tweak the error message and documentation — I'll try to steer people away from `g:slime_cell_delimeter`.

mroavi:
What about just echoing a deprecation message for some time. Example:
```
:Gstatus q is deprecated in favor of gq or the built-in <C-W>q
```
and at some point, when hopefully everybody adapted their config files, you drop `g:slime_cell_delimeter`

jpalardy:
Setting a `g:` variable is relatively straightforward; it can be done anywhere (but probably in .vimrc). As for `b:` variables, they imply a certain degree of comfort with vim that's hard to expect (or demand).

For now, I'm going to keep things as is.

If you can confirm that this PR works for you, I'll merge this.

mroavi:
This works for me. Thanks.

Issue #253, opened 2020-06-28
X11 Option Documentation: xdotool dependency and send-event flag


Jacob-Bishop:
Hello,
I had some issues getting the x11 option for vim-slime to work, and would like to suggest an addition to the README to make it easier for anyone else who runs into the same problem.
1) Sending text to another X window is dependent on xdotool. If it's not installed, then SlimeConfig appears to silently fail when ```g:slime_target = 'x11'``` (or at least ```:messages``` didn't show anything). It would be good to add a note about this to the README.
2) X11 servers set a special flag on text sent directly to a window using ```xdotool type```. Some terminals ignore text flagged this way by default (for example, termite does). See the ```SENDEVENT NOTES``` section of the xdotool man page. Again, a note about this may help someone else trying to trouble-shoot issues with the x11 option.
Thanks.

jpalardy:
Hi @Jacob-Bishop,

Good points!

I didn't write the x11 code and don't use it myself… so it's hard to determine what's obvious (or not) only from a PR.

If you want to add to the x11 section, I would be happy to merge that in. Likewise, if you think the first problem could (easily?) be flagged in messages, that might be a nice addition 😀

PR #261, opened 2020-10-05
Readme x11 target updates

Jacob-Bishop:
Hello, 
This pull request updates the readme to provide pointers for two stumbling blocks I ran into when setting the X11 target. This is the belated followup to issue #253. Let me know if you have any questions or suggestions.
Thanks,
Jacob


jpalardy:
Hi @Jacob-Bishop,

This is looking great, thanks for the PR 👍

PR #263, opened 2020-10-06
Fix #251

### LINKED ISSUE #251: (Neovim) Can't send raw data to rpc channel
I believe another [user](https://github.com/jpalardy/vim-slime/issues/212) had the same problem, but I do not know how they fixed it. I have just set up vim-slime in neovim and it is giving me an error of "can't send raw data to rpc channel" when I use the default shortcut(Control -C -C). It also asks me to set up jobid the first time I press the shortcut and automatically puts the number of 3 there. After that, this error just continues to show

image

HTransistor:
As in the posts to issue #251 explained, everytime when neovim terminal is called it gets a different job_id, so this one needs to be catched and directed to vim-slime somehow.
I used a global variable for this g:slime_jobid.
Importantly enough to get it automated the postfix:
<C-\><C-n> :let g:slime_jobid=b:terminal_job_id <CR>A
needs to be add to all commands that open the neovim terminal,
so the variable gets automatically assigned when opening a terminal.

The first time you need to accept and save the jobid by clicking enter.
It is the intended behaviour that opening a 2nd terminal, the old jobid stays and the text is still redirected to the old terminal.
To change this :SlimeConfig needs to be called with the new jobid, which is basically unkown,
so it is probably a good idea to assign <C-\><C-n> :let g:slime_jobid=b:terminal_job_id <CR>A to a short-cut in init.vim.

All the needed information for the user is added in the README.md.

### DIFF
--- a/autoload/slime.vim
+++ b/autoload/slime.vim
@@ -107,7 +107,7 @@ endfunction
 
 function! s:NeovimConfig() abort
   if !exists("b:slime_config")
-    let b:slime_config = {"jobid": "3"}
+    let b:slime_config = {"jobid": g:slime_jobid}
   end
   let b:slime_config["jobid"] = input("jobid: ", b:slime_config["jobid"])
 endfunction
--- a/README.md
+++ b/README.md
@@ -264,8 +264,10 @@ All commands to open nvim terminal need the following postfix: <C-\><C-n> :let g
 e.g. A-n to open terminal and get terminal_job_id for slime, then get into insert mode:  
 nnoremap <A-n> :call OpenTerminal() <CR><C-\><C-n> :let g:slime_jobid=b:terminal_job_id <CR>A  
 
-When you invoke vim-slime for the first time, you will be prompted for more
-configuration.
+The first time you need to accept and save the jobid by clicking enter.  
+It is the intended behaviour that opening a 2nd terminal, the old jobid stays and the text is still redirected to the old terminal.  
+To change this :SlimeConfig needs to be called with the new jobid, which is basically unkown,  
+so it is probably a good idea to assign <C-\><C-n> :let g:slime_jobid=b:terminal_job_id <CR>A to a short-cut in init.vim.  
 
 Advanced Configuration
 ----------------------
--- a/README.md
+++ b/README.md
@@ -259,6 +259,11 @@ for possible options, see :help term_start()
 
     let g:slime_target = "neovim"
 
+In ~/.config/nvim/init.vim:  
+All commands to open nvim terminal need the following postfix: <C-\><C-n> :let g:slime_jobid=b:terminal_job_id <CR>A  
+e.g. A-n to open terminal and get terminal_job_id for slime, then get into insert mode:  
+nnoremap <A-n> :call OpenTerminal() <CR><C-\><C-n> :let g:slime_jobid=b:terminal_job_id <CR>A  
+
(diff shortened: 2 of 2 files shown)


autoload/slime.vim
```diff
@@ -107,7 +107,7 @@ endfunction
 
 function! s:NeovimConfig() abort
   if !exists("b:slime_config")
-    let b:slime_config = {"jobid": "3"}
+    let b:slime_config = {"jobid": g:slime_jobid}
```
jpalardy, line 110: what's the behavior here if `g:slime_jobid` is undefined?

HTransistor: You get a prompt:

> Error detected while processing function slime#send_op[1]..<SNR>124_SlimeGetConfig[14]..<SNR>124_SlimeDispatch[6]..<SNR>124_NeovimConfig:                                                                                            
> line    2:
> E121: Undefined variable: g:slime_jobid

After enter:

> Error detected while processing function slime#send_op[1]..<SNR>124_SlimeGetConfig[14]..<SNR>124_SlimeDispatch[6]..<SNR>124_NeovimConfig:
> line    2:
> E15: Invalid expression: {"jobid": g:slime_jobid}

Clearly indicating you should read the README again...

jpalardy: hi @HTransistor,

although I agree the previous behavior (default to magic number `3`) wasn't great, I don't think this is a friendly workaround.

I think NeovimConfig could use a completion function like the Screen and Tmux integrations… but since I'm not familiar with the Neovim mechanism about assigning (and enquiring about) jobids, I'm not in a good position to write it.

HTransistor: I don't know why you think, that leaving it broken for everybody who want to use neovim, is better, but okay.

carlocab:
You can automate the setting of `g:slime_jobid` with something like
```nvim
augroup nvim_slime
  autocmd!
  autocmd TermOpen * let g:slime_jobid = b:terminal_job_id
augroup END
```
Of course, this means that `g:slime_jobid` will change every time the user opens a new terminal window, but since (I think) `NeovimConfig` will only query `g:slime_jobid` only once per buffer, that shouldn't be too big a problem. That is, unless you decide you want to send code snippets to a different terminal window than the one you started with, but I think that is true under the current state of the code anyway.

jpalardy:
hi @carlocab 

That sounds like a great solution, a few thoughts:
- maybe it could be named `g:slime_latest_jobid` (more explicit?)
- README needs to tell user to put this snippet in their vim config
- plugin code needs to catch unset `g:slime_latest_jobid`, and probably default to blank
- maybe README needs a blurb about `echo  &channel` or `echo b:terminal_job_id` to allow users to self-serve (when things go bad)

how does that sound?

carlocab:
Oh whoops, nvim lists `b:terminal_job_id` as deprecated under (see `:help b:terminal_job_id`). I think the `autocmd` should be `let g:slime_latest_jobid = &channel` instead.

jpalardy:
hi @carlocab 

good point about `b:terminal_job_id`…

I also thought we might make the `autocmd` conditional on `let g:slime_target = "neovim"`, which would "just work".

@HTransistor do you want to take this one? (if not, I'll open another PR … but I'll need someone to test it more thoroughly)

AkechiShiro:
I'll be glad to test the PR no matter who opens one up @jpalardy.

jpalardy:
closing, refer to #266

PR #264, opened 2020-10-15
Strip more comments from stata code lines

pacbard:
The current stata code only wraps lines that terminate with `///`.

Stata allows for 3 more types of comments:
 * `//` starts a comment anywhere on a line until the new line character
 * `/* */` bookends blocks of code (similar to c++ codeblocks)
 * `*` starts a comment code at the beginning of a line

I add support for the first two other types of code blocks in this pull request. . I don't add code for the third type of code block is not a problem because the Stata REPL treats it as a comment already. Adding it shouldn't be a problem if interested.

I tested with a simple do file and the code seems to be right, but I haven't written Vimscript before so I am not sure if it will work all the time. I will continue to test in the coming days.

### DIFF
--- a/ftplugin/stata/slime.vim
+++ b/ftplugin/stata/slime.vim
@@ -1,4 +1,6 @@
 function! _EscapeText_stata(text)
 	let remove_comments = substitute(a:text, '///\s*\n', " ", "g")
+	let remove_comments = substitute(remove_comments, '//.*\n', "\n", "g")
+	let remove_comments = substitute(remove_comments, '/\*.*\*/', "", "g")
 	return remove_comments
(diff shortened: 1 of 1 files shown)


jpalardy:
hi @pacbard,

It's looking good. Give it a few days and if it's still working well, let me know and I'll merge it.

Can I assume you'll be willing to help if problems come up later?

pacbard:
I have been using the updated code for a couple of days. Everything seems to work fine on my end.

I will be happy to follow up on any errors with the stata plugin that might come up later.

jpalardy:
Done, thank you ✅

Issue #194, opened 2019-03-27
Feature request - Allow selection to be sent as single line


mitermayer:
**Background**

I really enjoy that when doing "<c-c><c-c>" slime sends all the block lines one by one to the repl. But some repls require the block to be sent at once.

**Poblem**

Some repls do not support multine and require all commands to be inlined.

**Proposal**

 Would be great if there was an option to send all the current selection inlined into a single line for repls that do not support multiline commands

jpalardy:
Hi there,

By default, slime sends a block as a "chunk", not line by line. That behavior defaults on the ftplugin though.

What kind of code and REPL are you using?

mitermayer:
@jpalardy,

We use "hack" (https://github.com/facebook/hhvm) at FB. When I attempt to send it to the repl it is currently sending line by line which is incompatible with what `hphpd` expects.

jpalardy:
If you could show me a minimal example, I would try to reproduce and debug it myself.

Specifically:
- your setup (tmux, bash, other relevant)
- how you run the REPL
- what code block to send to reproduce

Thanks

avonmoll:
I'm not sure if this is actually the same issue that OP is describing, but I am seeing something like the following:

`.slime_paste` contains
```julia
x = 1

y = 2
```

and when this gets sent to the Julia REPL, for example, each line is evaluated individually:

```
julia> x = 1
1

julia>

julia> y = 2
2
```

However, inside Julia's REPL, you can enter multi-line commands by typing `x = 1`<kbd>Alt+Enter</kbd><kbd>Alt+Enter</kbd>`y = 2`<kbd>Enter</kbd>.

jpalardy:
Hi @avonmoll,

Trimming the blank line(s) between code lines is something that could be done by the ftplugin.

As for inserting <kbd>Alt-Enter</kbd> … that's trickier. Is <kbd>Alt-Enter</kbd> even a sequence that could be inserted in `.slime_paste` or something the REPL captures directly from its inputs 🤔

avonmoll:
That's an interesting question that I don't know the answer to. I think that where this issue might matter is in cases like the following

```julia
function multiplyBy2(x)
    return 2*x
end
```

[...]

PR #265, opened 2020-10-22
Use tmux bracketed paste

### LINKED ISSUE #262: vim-slime seems to append a "]" character to julia list comprehension lines
I just started using your useful plugin. Thank you for your nice work. 

I am using vim-slime to send julia code form nvim to julia REPL in tmux. 

in my .vimrc I have the following vim-slime config:

let g:slime_target = "tmux"                                                                            
let g:slime_default_config = {"socket_name": "default", "target_pane": "0.1"}

in nvim:
a = [ x for x in (1:1:10) ]  <C-c C-c> 

julia window:
julia> a = [ x for x in (1:1:10) ]]
ERROR: syntax: extra token "]" after end of expression  

note the extra square bracket added to the line as it was passed by vim-slime.

Similarly, if an array is declared, an extra "]" gets appended. 
in nvim:
ar = [1,2,3] <C-c C-c>

julia window:
julia> ar = [1,2,3]]
ERROR: syntax: extra token "]" after end of expression

However, the ipython REPL works exactly as expected. No extra brackets are added on.

VERSIONS:
julia version 1.5.2

NVIM v0.5.0-nightly
Build type: Release
LuaJIT 2.0.5

tmux 3.1b

no version is provided in vim-slime (but installed from github on ~Sept 30, 2020)

carlocab:
Fixes #262, https://github.com/KristofferC/OhMyREPL.jl/issues/224, https://github.com/jpalardy/vim-slime/issues/194#issuecomment-517429613
Related to #254

I tried looking for terminal applications that don't support bracketed paste https://github.com/jpalardy/vim-slime/pull/254#issuecomment-694724480 but haven't managed to find any. Playing around a little bit with it Python hasn't revealed any problems either.

### DIFF
--- a/autoload/slime.vim
+++ b/autoload/slime.vim
@@ -74,7 +74,7 @@ endfunction
 function! s:TmuxSend(config, text)
   call s:WritePasteFile(a:text)
   call s:TmuxCommand(a:config, "load-buffer " . g:slime_paste_file)
-  call s:TmuxCommand(a:config, "paste-buffer -d -t " . shellescape(a:config["target_pane"]))
+  call s:TmuxCommand(a:config, "paste-buffer -p -d -t " . shellescape(a:config["target_pane"]))
 endfunction
 
 function! s:TmuxPaneNames(A,L,P)


jpalardy:
hi @carlocab 

I've been running this patch (manually) for a few weeks now, I don't see bad side-effects.

Let's merge it for now, and see what comes up.

carlocab:
Great! If it helps allay any worries you have, tmux is soon making bracketed-paste the default behaviour for `<Prefix>-]`:

https://github.com/tmux/tmux/blob/9a5b2c93a7129a6e6e74e4b6d4c2c78dd568b8d4/CHANGES#L13

kdauria:
I don't quite understand why, but this change broke my usual workflow. Within a tmux session inside of an ubuntu20.04 docker container, I send commands from vim to ipython. With the latest commit (`dein#add('jpalardy/vim-slime')`)

Screen Shot 2020-10-24 at 9 54 14 AM

When I manually deleted the two dashes causing the syntax error and then press Enter to issue the commands, the ipython terminal just hangs forever.

With an older commit (`dein#add('jpalardy/vim-slime', { 'rev': 'e7b3cb0bb55a6ca65b218d7e825d9aac71d60638' })`)

Screen Shot 2020-10-24 at 9 47 36 AM

It seems like with this update that the `cpaste -q` is no longer needed?

carlocab:
@kdauria Ugh, sorry about that. What happens if you remove `%cpaste -q` and `--` entirely?

I don't use IPython much, but from my (admittedly limited, and possibly mistaken) understanding is that `%paste` and `%cpaste` were early implementations of bracketed-paste before terminal apps started supporting it. I suspect you should now be able to get by without `%cpaste` with the new update, but unfortunately, it will also break any old code you have.

kdauria:
Thanks @carlocab . I'll try fiddling around with this line: https://github.com/jpalardy/vim-slime/blob/92421f0da36193a2b4dda37b8e9efbf278e93bf5/ftplugin/python/slime.vim#L8

carlocab:
> Thanks @carlocab . I'll try fiddling around with this line:
> 
> https://github.com/jpalardy/vim-slime/blob/92421f0da36193a2b4dda37b8e9efbf278e93bf5/ftplugin/python/slime.vim#L8

@kdauria 

Yes, I think that should do it! I think you'll want to keep the `\n` at the end, but with that, your snippet should execute fine. I'm not sure how it'll handle indentation though.

If you have a code snippet with indentation I could play with to see what works, please do share it so I can try to test it out too.

carlocab:
Ok, now that I think about it playing around with that line will probably affect users who aren't using tmux.

It might be better to just deactivate bracketed paste if `g:slime_python_ipython = 1`.

kdauria:
@carlocab I gave it a try. https://github.com/kdauria/vim-slime/pull/1

I'm stuck on how to send a carriage return in ipython in order to "submit" the commands. I tried adding <CR> in my vim config, but that didn't work.

```
    xmap <Leader>e <Plug>SlimeRegionSend<CR>
    nmap <Leader>e <Plug>SlimeParagraphSend<CR>
```

FWIW, I really like the bracketed paste. It seems to be less buggy and also allows ipython to do syntax highlighting.

carlocab:
Yea, I got stuck on the same thing. I tried looking up how to get IPython to read a character as a carriage return, but no luck.

carlocab:
A potential, but possibly ugly, solution would be for `TmuxSend` not to invoke the `-p` flag, but to wrap the text it sends with the bracketed-paste escape codes directly. This would get around the problem of bracketed-paste eating the `\n` after `%cpaste` and `--`, and should in theory exhibit identical behaviour elsewhere.

However, I don't know what `%cpaste` does with those escape codes. I'll try to carve out time tomorrow to try it out.

jpalardy:
reverted 10e8c2b3bd19e99c4889e4135a34eaf5aa0b37c3

PR #266, opened 2020-10-24
Use `g:slime_last_channel` and autocmd to get a better jobid for neovim

jpalardy:
Addresses #251 #263

### DIFF
--- a/README.md
+++ b/README.md
@@ -260,7 +260,14 @@ for possible options, see :help term_start()
     let g:slime_target = "neovim"
 
 When you invoke vim-slime for the first time, you will be prompted for more
-configuration.
+configuration. The last terminal you opened before calling vim-slime will
+determine which `job-id` is presented as default.
+
+To manually check the right value of `job-id` to use, try:
+
+    echo &channel
+
+from the buffer running your terminal.
 
 Advanced Configuration
 ----------------------
--- a/autoload/slime.vim
+++ b/autoload/slime.vim
@@ -107,7 +107,7 @@ endfunction
 
 function! s:NeovimConfig() abort
   if !exists("b:slime_config")
-    let b:slime_config = {"jobid": "3"}
+    let b:slime_config = {"jobid": get(g:, "slime_last_channel", "")}
   end
   let b:slime_config["jobid"] = input("jobid: ", b:slime_config["jobid"])
 endfunction
--- a/plugin/slime.vim
+++ b/plugin/slime.vim
@@ -36,3 +36,11 @@ if !exists("g:slime_no_mappings") || !g:slime_no_mappings
   endif
 endif
 
+" for neovim (only), make slime_last_channel contain
+" the channel id of the last opened terminal
+if get(g:, "slime_target", "") == "neovim"
+  augroup nvim_slime
+    autocmd!
+    autocmd TermOpen * let g:slime_last_channel = &channel
+  augroup END
+end


jpalardy:
cc: @kujbika @AkechiShiro @a-elhag 

if you can give this a try and let me know if it works for you 😄

kujbika:
Cant seem to work for me ("Invalid job id"), however I might mess something up.

carlocab:
Works fine for me. Though I also noticed that it doesn't seem to use bracketed paste... :) (Just as an FYI, in case it comes up later on.)

AkechiShiro:
I tested it it works but there is a slight issue, if the first terminal launched is quit in vim then the jobid is not updated for the new terminal launched it throws an `invalid channel id` error.

I don't know if this can be fixed but in my configuration when a terminal is quit, and then a new one is spawned in Neovim, the new terminal `job_id` is the `old_term_job_id + 1` so in my case I believe it shouldn't be hard to fix this minor issue.

carlocab:
Hey @AkechiShiro, this is [anticipated behaviour](https://github.com/jpalardy/vim-slime/pull/263#issuecomment-715464716). Though I suppose it could be better documented.

Would you know if there is neovim documentation to support that new terminals have a `job_id` incremented by one?

AkechiShiro:
I don't really know of neovim documentation that supports my claim but looking at the documentation about [`job_control`](https://neovim.io/doc/user/job_control.html#job-id) I don't see anything mentioning this. @carlocab

carlocab:
Oh ok. I'm having a look at the docs too. I'll send an update here if I find something useful.

jpalardy:
I hadn't _really_ played with neovim before this PR…

Here's some notes from my experiments:
- every opened terminal updates `g:slime_last_channel`, but it's "locked in" the first time slime is invoked
- if the terminal is closed, slime is not aware of that (currently, I guess there could be events for this 🤔 )
- opening a new terminal doesn't change the slime config either (buffer still locked on previous config)

Other modes (screen, tmux, etc) fail silently, so this isn't too different.

I agree the documentation could reflect this behavior. I'm also open to improvements if you have suggestions.

carlocab:
> * if the terminal is closed, slime is not aware of that (currently, I guess there could be events for this 🤔 )

Yes, there is the event `TermClose`. I guess one could write an `autocmd` that resets `b:slime_config` when this happens. However, I think doing it properly could prove quite tricky.

jpalardy:
> I think doing it properly could prove quite tricky.

same here, especially since the `b:slime_config` lives in a different buffer than where the event is triggered (or, worse, that multiple buffers might be affected 😱 )

carlocab:
I guess one could check for the existence of `b:slime_config`, and whether `b:slime_config["jobid"] = &channel` (hopefully `&channel` is still set correctly when this event is triggered...). If both those conditions are true it should be safe to reset `b:slime_config`. I don't know how this will work for the case where there are multiple buffers sending text to the same terminal window, however, and it probably doesn't work as smoothly as one would like when there are other, unrelated buffers open as well.

jpalardy:
@kujbika 

would you mind describing the flow that doesn't work for you?

kujbika:
Hi, I checked once again, and it works properly now. Thanks for the fix, highly appreciated!

jpalardy:
Let's merge it and keep an eye on it.

Let's iterate on this: please open another issue if it's not working or if we can make it better.

PR #278, opened 2020-12-31
Update documentation for slime_paste_file

pfheatwole:
Using STDIN was deprecated in tmux 2.2


jpalardy:
Looking good, thank you!

Issue #279, opened 2021-01-06
vim-slime doesn't work with kitty


Svzer:
Hi !

I've recently started to use vim-slime with tmux and it worked flawlessly, but with tmux being unable to display images in the terminal, I wanted to try and rely solely on kitty. However, I just can't seem to make it work : every time I enter the target window id, nothing happens. It does seem like the content is being passed to the .slime_paste file, but it is never sent to the target window.

I would gladly appreciate some help. Thanks in advance !

jpalardy:
Hi @Svzer 

I've also seen weird behavior while trying to display images in iterm2, using tmux. (sometimes it works, sometimes it doesn't … something it doesn't leave enough space for the whole height)

The kitty integration is super simple:

![Screen Shot 2021-01-06 at 09 16 38](https://user-images.githubusercontent.com/18142/103799429-ee684e00-4fff-11eb-890a-e5becec2de20.png)

Can you try to:
- create a file and put content in it (or re-use the slime_paste file)
- manually run the command from line 49 manually and see what happens

Let me know how that goes and we'll take it from there.

Svzer:
Hi,

I'm sorry I forgot to mention this : I already tried manually running this command, and it worked.
I'm very much a novice, but is there a need for the slime_paste file for the kitty integration though ? kitty is able to read text sent from stdin, so I thought perhaps this could be a workaround.

jpalardy:
Hi @Svzer 

I installed kitty and … it doesn't seem to work for me (even manually)

Can you give me a step-by-step?
- which version you're running
- what's in your kitty.conf (saw that allow_remote_control might be important)
- what command you tried — and which ones worked?

If I can reproduce what works/doesn't, that would help 😄

Svzer:
Hi,

I'm running the **0.19.3** version.
Here's my kitty configuration : 
```
# Color theme
include snazzy.conf

font_family	MesloLGSDZ Nerd Font
font_size	12
window_margin_width 0 10

# Tabs
tab_bar_style powerline
active_tab_background #5AF78E
active_tab_font_style bold
inactive_tab_font_style normal

# Windows
active_border_color #5AF78E
window_border_width 2px
window_margin_width 2
window_padding_width 4

# Allow remote control for vim-slime to work
allow_remote_control yes
```
What I've done is : 

- Open a new kitty terminal
- Open a new window inside that terminal with Ctrl+Shift+Enter
[...]

Issue #234, opened 2020-04-03
feature request: kitty @ --to


jbrea:
Hi

I love vim-slime, thank a lot!
Currently I use it with kitty.
Recently I realised that there is a second way to do [remote control in kitty](https://sw.kovidgoyal.net/kitty/remote-control.html#tutorial).
To try it, I changed the KittySend function to
```
  call system("kitty @ --to unix:/tmp/" . shellescape(a:config["window_id"]) .
    \  " send-text --from-file " . g:slime_paste_file)
```
This works. But what is the best way to incorporate both ways of using kitty remote control into vim-slime?

jpalardy:
hi @jbrea,

This is similar (though not exactly the same) to the current code.

What's the advantage of this approach? (or of supporting both?)

jbrea:
With the current approach messaging works only when run inside a kitty window.
With the approach I proposed above one can control kitty also from programs/scripts not running inside a kitty window. But this increased flexibility comes at the cost of having to start kitty like this
`kitty -o allow_remote_control=yes --listen-on unix:/tmp/mykitty`.

My initial motivation for the new approach was to send text to a kitty window that runs in a different OS window. But now I realized that one can detach kitty windows into new OS windows. So I am fine with the current approach of vim-slime.

I am closing this for now, if you don't mind.

PR #280, opened 2021-01-07
Use alternate Kitty text send method supporting Neovim

mdzhang:
## Summary

Leverages suggestion from https://github.com/jpalardy/vim-slime/issues/234

in response to https://github.com/jpalardy/vim-slime/issues/279

## Testing

Manually confirm slime works when inside a Kitty window (0.18.3) started with documented options, inside regular vim (8.2) and neovim 0.4.4

### DIFF
--- a/README.md
+++ b/README.md
@@ -183,6 +183,25 @@ kitty target window
     This is the id of the kitty window that you wish to target.
     See e.g. the value of $KITTY_WINDOW_ID in the target window.
 
+kitty listen on
+
+    Specifies where kitty should listen to control messages.
+    See e.g. the value of $KITTY_LISTEN_ON in the target window.
+
+To work properly, `kitty` must also be started with the following options:
+
+```sh
+kitty -o allow_remote_control=yes -o enabled_layouts=tall --listen-on unix:/tmp/mykitty
+```
+
+See more [here](https://sw.kovidgoyal.net/kitty/remote-control.html). This can also be added to your `kitty.conf` file as:
+
+```
+allow_remote_control yes
+enabled_layouts tall
+listen_on unix:/tmp/mykitty
+```
+
 ### X11
 
 [x11](http://manpages.ubuntu.com/manpages/trusty/man1/xdotool.1.html) is *not* the default, to use it you will have to add this line to your
@@ -277,7 +296,7 @@ from the buffer running your terminal.
 Advanced Configuration
 ----------------------
 
-If you need this, you might as well refer to [the code](https://github.com/jpalardy/vim-slime/blob/master/plugin/slime.vim#L233-L245) 😄  
+If you need this, you might as well refer to [the code](https://github.com/jpalardy/vim-slime/blob/master/plugin/slime.vim#L233-L245) 😄
 Seriously, it's not a complicated as it seems.
 
 If you don't want the default key mappings, set:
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -183,6 +183,25 @@ kitty target window
...
 
+kitty listen on
+
+    Specifies where kitty should listen to control messages.
+    See e.g. the value of $KITTY_LISTEN_ON in the target window.
+
+To work properly, `kitty` must also be started with the following options:
+
+```sh
+kitty -o allow_remote_control=yes -o enabled_layouts=tall --listen-on unix:/tmp/mykitty
```
jpalardy, line 194: is `-o enabled_layouts=tall` necessary?
(same below @ line 201)

README.md
```diff
@@ -277,7 +296,7 @@ from the buffer running your terminal.
 Advanced Configuration
 ----------------------
 
-If you need this, you might as well refer to [the code](https://github.com/jpalardy/vim-slime/blob/master/plugin/slime.vim#L233-L245) 😄
```
jpalardy, line 280: this whitespace causes a `<br/>`
(if you could revert that line)

autoload/slime.vim
```diff
@@ -46,15 +46,17 @@ endfunction
...
-  call system("kitty @ send-text --match id:" . shellescape(a:config["window_id"]) .
+  call system("kitty @ --to " . shellescape(a:config["listen_on"]) .
+    \ " send-text --match id:" . shellescape(a:config["window_id"]) .
     \ " --from-file " . g:slime_paste_file)
 endfunction
 
 function! s:KittyConfig() abort
   if !exists("b:slime_config")
-    let b:slime_config = {"window_id": 1}
+    let b:slime_config = {"window_id": 1, "listen_on": $KITTY_LISTEN_ON}
```
jpalardy, line 56: I have mixed feelings about this … now everybody needs to specify a `listen_on` 😩 

maybe we could leave it blank and skip the `--to` when it's blank? (is that better or more confusing?)

jpalardy (commented):
excellent, but I would like @svzer to confirm 👍

README.md
Svzer: It is very likely not needed here

autoload/slime.vim
Svzer: Would it work to simply add a `if has('nvim')` condition in order not to force regular vim users to specify a listen_on for vim-slime to work in kitty ?

mdzhang: i guess if you had multiple kitty sessions open, and wanted to slime it to a different session, the override would be helpful, but otherwise hitting enter through the second input is what i think most people would do

if we decide that overriding is sufficiently unlikely that we just default to `KITTY_LISTEN_ON`, in the case where you're in nvim and `$KITTY_LISTEN_ON` is empty it could also just error

at any rate i put the nvim conditional on it out of laziness, but am open to follow ups if folx feel strongly about it

jpalardy: I got burned myself on:
- missing `allow_remote_control=yes`
- calling `kitty @ send-text` from iterm2 ("outside" kitty)

so, I don't think it's just an nvim problem.

I think this PR makes things better, but I'll add to it

jpalardy:
integrated in #281, thanks

PR #281, opened 2021-01-08
Modifications to PR #280

jpalardy:
cc: @mdzhang @Svzer

I did a few changes to #280 
- use stdin: no more paste file
- uses `--to` only if `listen_on` isn't blank
- no special nvim cases
- more notes to README

### DIFF
--- a/README.md
+++ b/README.md
@@ -183,6 +183,25 @@ kitty target window
     This is the id of the kitty window that you wish to target.
     See e.g. the value of $KITTY_WINDOW_ID in the target window.
 
+kitty listen on
+
+    Specifies where kitty should listen to control messages.
+    See e.g. the value of $KITTY_LISTEN_ON in the target window.
+
+To work properly, `kitty` must also be started with the following options:
+
+```sh
+kitty -o allow_remote_control=yes -o enabled_layouts=tall --listen-on unix:/tmp/mykitty
+```
+
+See more [here](https://sw.kovidgoyal.net/kitty/remote-control.html). This can also be added to your `kitty.conf` file as:
+
+```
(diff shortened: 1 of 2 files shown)


jpalardy:
found a funny case, and _tried_ to explain it in the README:
- if KITTY_LISTEN_ON is defined/exported
- kitty doesn't need the `--to` flag at all…  💥 
- so `listen_on` can be left blank…

alternatively, and less magically, I used:

```
let g:slime_target = "kitty"
let g:slime_default_config = {"window_id": 1, "listen_on": $KITTY_LISTEN_ON}
```

so the fields are pre-filled. If that's useful, we could patch the README

jpalardy:
things to try:
- inside/outside kitty
- explicit/blank `listen_on`
- vim/nvim

I did a quick pass, and it seems to work. I would be grateful if you could give it a quick pass.

Svzer:
Works fine for me on both vim and nvim, although I'm never prompted to enter a `listen_on` in nvim. 

I don't plan to use it, but I could send text from alacritty to kitty by manually entering the listen_on. Only worked with vim though, with nvim I'm not asked to enter a listen_on and get stuck.

jpalardy:
so weird … was it running the right version of the plugin? (that's the only thing I can think of…)

Svzer:
That was indeed a mistake on my part after I double-checked, everything works fine, even with nvim. Sorry about that.

jpalardy:
excellent, thanks ^

jpalardy:
let's merge this for now, if you can make the README clearer, I would merge that too.

thanks: @Svzer @mdzhang

PR #302, opened 2021-11-03
udpate for kitty to select window

xiangpeng2008:
Kitty added a very cool new feature to select window to send command
We don't need to enter manually window ID, which is not evident to find, instead, kitty will pop up numbers for us to choose.
https://github.com/kovidgoyal/kitty/issues/4165#issue-1039321261 
vim_slime_demo

### DIFF
--- a/autoload/slime.vim
+++ b/autoload/slime.vim
@@ -56,7 +56,7 @@ function! s:KittyConfig() abort
   if !exists("b:slime_config")
     let b:slime_config = {"window_id": 1, "listen_on": ""}
   end
-  let b:slime_config["window_id"] = input("kitty target window: ", b:slime_config["window_id"])
+  let b:slime_config["window_id"] = str2nr(system("kitty @ select-window --self"))
   let b:slime_config["listen_on"] = input("kitty listen on: ", b:slime_config["listen_on"])
 endfunction
 


jpalardy:
Hi @xiangpeng2008 

This is looking good!

I'll merge this and `@` you if anything comes up.

euglevi:
But is this already implemented in kitty or is it a new feature that is not incorporated into the stable version yet? I cannot seem to find a way to make it work with the 0.23.1 version of kitty.

jpalardy:
@xiangpeng2008 thoughts?

xiangpeng2008:
hi @euglevi 
It's at least integrated in master branch, if the last stable release doesn't have this, please try the night build https://github.com/kovidgoyal/kitty/issues/3948

euglevi:
Ok I have tried it. The select-window works but SlimeConfig asks me the listen_on value: is this supposed to be the case under nvim? I was under the impression that the issue was solved with pull #280. Then, with the nightly build the vim-kitty-navigator plugin seems to be broken, even tough this issue is clearly not related to vim-slime.

xiangpeng2008:
glad to know that select-window works for you now.
I always use "vim" so I'm not aware of your issue with nvim. Sorry about that.

Actually I'm not sure if "listen_on" serves something in "vim" so I just set it to always "" in my personal branch.

```
let b:slime_config["listen_on"] = ""
```

Please make a request to Kovid if you think the broken vim-kitty-navigator is related to kitty.

euglevi:
I got this to work! Vim-kitty-navigator at the moment is broken with me but vim-slime is not.

euglevi:
The selection through the numbers on the screen broke for me with the last updates of vim-slime. I am not sure why. Any ideas?

jpalardy:
hi @euglevi 

can you pinpoint what commit broke it for you?

euglevi:
Weird... it seems that it was related to the Kitty's theme I was using so not a vim-slime problem after all. I have to say that I do not have time to debug this but, again, it does not look like a vim-slime issue at all.

jpalardy:
keep me posted 👍

Issue #49, opened 2014-03-13
vim tmux error


sevakd:
"Error detected while processing function <SNR>10_SlimeSendOp..<SNR>10_SlimeSend..<SNR>10_SlimeDispatch..
<SNR>10_TmuxSend:
line    4:
E484: Can't open file /tmp/vrhkGsT/1"

Unsure if I missed a step as far as setting things up.
When I go to that directory within /tmp/ there is a file named 0, but not 1.
Don't know if that is of any help.

kariya2:
@sevakd  facing the same issue where I'm unable to open the tmp file. what helped you fix it?

blackheaven:
I have the same issue, any hints?

jpalardy:
As reported by @blackheaven in #307 

```vim
let g:slime_paste_file = tempname()
au VimLeave * call delete(g:slime_paste_file)
```

can solve this problem and cleanup

PR #307, opened 2021-12-12
Use a temporary file for slime_paste_file

blackheaven:
I encountered #49

It occurs that creating a dedicated temporary file seems sound anyway.

### DIFF
--- a/autoload/slime.vim
+++ b/autoload/slime.vim
@@ -12,7 +12,8 @@ end
 
 " screen and tmux need a file, so set a default if not configured
 if !exists("g:slime_paste_file")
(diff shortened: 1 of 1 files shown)


jpalardy:
Hi @blackheaven 

I've opted not to make this the default. The [README](https://github.com/jpalardy/vim-slime) (look for `tempname`) discusses a solution similar to yours.

I like this part though:

```vim
au VimLeave * call delete(g:slime_paste_file)
```

and I would recommend putting it in your configuration rather than in the plugin.

I'll add a comment on #49 

Thanks

PR #308, opened 2021-12-19
unset `slime_config` for automatically configuration next time

you-n-g:
If we fail to successfully set the right value for `slime_config` in neovim terminal for the first time, `slime_config`  will be set to some value like  `{"jobid": ""}`.
Then it will not leverage `g:slime_last_channel` to automatically config `slime_config`  next time.

So we unlet `slime_config`  if it fails.


jpalardy:
Hi @you-n-g 

That makes sense and it's looking good 👍 

Thanks for pitching in 😄

Issue #309, opened 2021-12-19
workaround till kitty select window is in main


ArbitRandomUser:
what i do ..
open two kitty's side by side
check output of $KITTY_LISTEN_ON in one of them
open nvim in other one (i have it aliased to vim if that matters) 

ctrl-c ctrl-c in nvim 
it prompts me for kitty listen on (but not window id)
i enter the $KITTY_LISTEN_ON that i saw in the other window
but i dont see any output slimed over .

neovim version 
```
NVIM v0.7.0-dev+662-g73b35ef10
Build type: RelWithDebInfo
LuaJIT 2.0.5
```

vim-slime version 
```
commit 7c49bd710e11268cac56cecfe6ff0426d0026808 (grafted, HEAD -> main, origin/main, origin/HEAD)
Author: you-n-g <you-n-g@users.noreply.github.com>
Date:   Mon Dec 20 00:16:24 2021 +0800

    unset `slime_config` for automatically configuration next time (#308)
lines 1-5/5 (END)
```

kitty version 
```
kitty 0.23.1 created by Kovid Goyal
```

ArbitRandomUser:
upong furthur inspection this seems to be cause by this line(52) in 
autoload/slime.vim
  ```call system("kitty @" . to_flag . " send-text --match id:" . shellescape(a:config["window_id"]) . " --stdin", a:text)```
  
changing this to 
```
call system("kitty @" . to_flag . " send-text  --stdin", a:text)     
```
seems to make it work as expected

ArbitRandomUser:
okay the real issue is kitty @ select-window is not yet implemented in 0.23.1
maybe dont implement select window till its out on a release branch/tag yet ?

ArbitRandomUser:
iv'e made a pull request , https://github.com/jpalardy/vim-slime/pull/310

jpalardy:
Merged ✅

PR #311, opened 2021-12-19
Arbitrandombranch

ArbitRandomUser:
ive made a small fix for issue #309 till kitty select window is available in a stable branch of kitty

### DIFF
--- a/autoload/slime.vim
+++ b/autoload/slime.vim
@@ -56,8 +56,12 @@ function! s:KittyConfig() abort
   if !exists("b:slime_config")
     let b:slime_config = {"window_id": 1, "listen_on": ""}
   end
-  let b:slime_config["window_id"] = str2nr(system("kitty @ select-window --self"))
   let b:slime_config["listen_on"] = input("kitty listen on: ", b:slime_config["listen_on"])
+  let b:slime_config["window_id"] = str2nr(system("kitty @ select-window --self"))
+  if v:shell_error
+    echom 'kitty @select-window failed setting id to 1'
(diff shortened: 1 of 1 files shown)


jpalardy:
Looks good to me ^

I might ping you depending on how things evolve

Thanks for the PR

ArbitRandomUser:
may be we should let the return of
`str2nr(system("kitty @ select-window --self"))` to a temp variable , and assign it later in an else 
like this ..
```
function! s:KittyConfig() abort                                                                                                                                                                
  if !exists("b:slime_config")                                                                                                                                                                 
  ¦ let b:slime_config = {"window_id": 1, "listen_on": ""}                                                                                                                                     
  end                                                                                                                                                                                          
  let l:temp = str2nr(system("kitty @ select-window --self"))                                                                                                                                  
  if v:shell_error                                                                                                                                                                             
  ¦ let b:slime_config["window_id"] = input("kitty window_id: ",b:slime_config["window_id"])                                                                                                   
  else                                                                                                                                                                                         
  ¦ let b:slime_config["window_id"] = l:temp                                                                                                                                                   
  end                                                                                                                                                                                          
  let b:slime_config["listen_on"] = input("kitty listen on: ", b:slime_config["listen_on"])                                                                                                    
endfunction                             
```
This way the `b:slime_config["window_id"]` , isn't polluted from the return value and the input prompt 
suggests the right value ,

jpalardy:
it doesn't seem like a big problem, but I see what you mean ^

I think I'm fine with either approaches 🤷

PR #314, opened 2021-12-26
A trade-off for #265, allowing to have bracketed-paste in tmux

Klafyvel:
I reused the work in #265 to allow enabling bracketed-paste in tmux if needed (it's useful in some cases for Julia's OhMyREPL.jl at least).

### DIFF
--- a/README.md
+++ b/README.md
@@ -148,25 +148,25 @@ Note that all of these ordinals are 0-indexed by default.
     "{token}" one of tmux's supported special tokens, like "{last}"
 
 
+You can configure the defaults for these options. If you generally run vim in
+a split tmux window with a REPL in the other pane:
+
+    let g:slime_default_config = {"socket_name": get(split($TMUX, ","), 0), "target_pane": ":.2"}
+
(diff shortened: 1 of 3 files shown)


jpalardy:
Hi @Klafyvel 

I like the idea, but not necessarily the configuration flag.

Putting it in `slime_config` and prompting the user seems obtrusive. (especially for something that wasn't in vim-slime for all theses years)

Maybe it could be a global (and buffer) flag, `g:slime_bracketed_paste`, with the following logic:
- use `b:slime_bracketed_paste` if found  --- set in ftplugin, presumably
- use `g:slime_bracketed_paste` if found

What do you think?

Klafyvel:
Hi, 
I think it's a good idea, to be honest I was unsure which was the cleanest way to add this configuration. I'll update the pull-request !

jpalardy:
I appreciate the PR, thanks for pitching in 👍

jpalardy:
Looking good ^

I'll merge this and might loop you in if this is causing problem.

Thanks again 🚀

PR #315, opened 2021-12-27
Add basic commands to navigate between cells

Klafyvel:
This adds three commands to teach slime how to navigate between cells.

### DIFF
--- a/README.md
+++ b/README.md
@@ -344,6 +344,9 @@ If you want to send blocks of code between two delimiters, emulating the cell-li
 
     let g:slime_cell_delimiter = "#%%"
     nmap <leader>s <Plug>SlimeSendCell
+    nmap <leader>sn <Plug>SlimeNextCell
+    nmap <leader>sp <Plug>SlimePrevCell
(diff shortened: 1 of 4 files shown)


jpalardy:
Hi @Klafyvel 

I'm unclear as to how much use this cell feature is getting… 🤔 

Since it lives "on top" of vim-slime, it makes me wonder if it could be a separate plugin.

What do you think?

Klafyvel:
Hi,
Although the cell feature is what drove me to vim-slime, it's true that it
feels a bit like an addon on top of slime.

I'm perfectly fine with making my own plugin for this feature, especially
since I have some other ideas to improve it. 🙂


Le mar. 28 déc. 2021 à 06:30, Jonathan Palardy ***@***.***> a
écrit :

Klafyvel:
Here is the plugin for future reference (and self-promotion :innocent: ) : https://github.com/Klafyvel/vim-slime-cells

jpalardy:
If you feel it's ready (or when you do), let's open a PR to link to it

At that point, we might consider stripping out any cell code from vim-slime and delegate 🤔

PR #316, opened 2021-12-29
Add a link to vim-slime-cells' github repo.

Klafyvel:
Following your proposal in https://github.com/jpalardy/vim-slime/issues/315#issuecomment-1002204207_ , I added a link towards my plugin. The new plugin does not use vim-slime's cell feature so you can remove it when you feel the time has come. 

Maybe it is interesting to keep it for now, in order not to break other users configuration ?


jpalardy:
Looking good — thank you 👍

PR #317, opened 2022-01-05
Made comment matching non-greedy

Eph97:
Made matching of comments non-greedy. Before when sending a block of code, it would match from the first `//` comment to the end of the last line. Also fixed matching so that comments at the end of line must be proceeded by whitespace as specified in stata documentation. This should also be a better fix for the problem in #313 because a true comment must be proceeded by whitespace.


jpalardy:
Hi @Eph97 

Can you look at that conflict?

Eph97:
Oops, fixed.

jpalardy:
Looking good, thanks ^

PR #319, opened 2022-01-13
Add :SlimeSend0 to vim-slime.txt

### LINKED ISSUE #318: :SlimeSend1 <C-d> sends (un)expected extra carriage return
I have `<Leader>SD` mapped to `<Cmd>SlimeSend1 <C-d><Cr>`. This properly exits the running REPL, but then sends an additional `<Cr>` resulting in the following shell:

image

Given the following help text:

```
:SlimeSend1 {text}      Send a single line of text, specified on the command
                        line, to screen, tmux, or whimrepl. A carriage return
                        is automatically appended.
```

The behavior actually makes sense, it's just kinda funky to end up with an empty shell prompt after the REPL is closed.

knpwrs:
Closes #318

### DIFF
--- a/doc/vim-slime.txt
+++ b/doc/vim-slime.txt
@@ -59,7 +59,8 @@ benefits of using Vim (familiar environment, syntax highlighting, persistence
 
                                                 *:SlimeSend0*
 :SlimeSend0 {text}      Send a single line of text, specified on the command
-                        line, to screen, tmux, or whimrepl.
+                        line, to screen, tmux, or whimrepl. This will not
+			automatically append a carriage return.
 
 
 ==============================================================================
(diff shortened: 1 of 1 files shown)


doc/vim-slime.txt
```diff
@@ -57,6 +57,10 @@ benefits of using Vim (familiar environment, syntax highlighting, persistence
                         line, to screen, tmux, or whimrepl. A carriage return
                         is automatically appended.
 
+                                                *:SlimeSend0*
+:SlimeSend0 {text}      Send a single line of text, specified on the command
+                        line, to screen, tmux, or whimrepl.
```
jpalardy, line 62: Would it be better to add something like "this won't send a carriage return" 🤔

knpwrs: Done in 708c244

jpalardy:
awesome, thank you 🚀

PR #328, opened 2022-05-07
Fix slime_bracketed_paste does not execute command

xltan:
This will send an Enter key to issue the command

### DIFF
--- a/autoload/slime.vim
+++ b/autoload/slime.vim
@@ -78,8 +78,6 @@ function! s:TmuxCommand(config, args)
 endfunction
 
 function! s:TmuxSend(config, text)
-  call s:WritePasteFile(a:text)
-  call s:TmuxCommand(a:config, "load-buffer " . g:slime_paste_file)
   if exists("b:slime_bracketed_paste")
     let bracketed_paste = b:slime_bracketed_paste
   elseif exists("g:slime_bracketed_paste")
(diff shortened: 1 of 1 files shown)


autoload/slime.vim
```diff
@@ -78,17 +78,28 @@ function! s:TmuxCommand(config, args)
...
+    let last_char_is_carriage_return = 1
+  else
+    call s:WritePasteFile(a:text)
+  end
+
+  call s:TmuxCommand(a:config, "load-buffer " . g:slime_paste_file)
   if bracketed_paste
     call s:TmuxCommand(a:config, "paste-buffer -d -p -t " . shellescape(a:config["target_pane"]))
+    if last_char_is_carriage_return
+      call s:TmuxCommand(a:config, "send-keys -t " . shellescape(a:config["target_pane"]) . " Enter")
```
jpalardy, line 101: is `Enter` different from … newline?
(if the `\r` had been a `\n`, would it have worked?)

jpalardy (commented):
Hi @xltan 

Thanks for the RP, one comment inline.

I'm trying to see if this can be simplified.

autoload/slime.vim
xltan: Just added some code to also handle `\n` and `\r\n`, so `SlimeRegionSend` will work.
The Enter in tmux is an emulated key, defined here as `\r`: https://github.com/tmux/tmux/blob/3b7dae9a5339c4495d30a4c1bb28ffe8aead0b27/key-string.c#L63

jpalardy: actually, if you could tell me how to reproduce both a positive and negative case — I could test this on my side too 👍

xltan: I think they didn't act that different, only `send-keys` and `paste` acts differently. This PR  issue a `send-keys enter` if the last char is crlf instead, send-keys will not be issued as [bracketed-paste](https://en.wikipedia.org/wiki/Bracketed-paste), so works just like you pressed Enter physically. From the wiki paste mode is using special terminal keys to quote the content, it will not autoindent, faster, almost always want we want, except it just paste the content, will never execute it even it's a newline.

jpalardy:
Hi @xltan 

I'm ready to merge this. As I mentioned, I'd like some scenarios to test it working, and not working (both before/after your PR)
- what REPL are you sending into?
- what are you sending? (if relevant)
- what works / doesn't work?

xltan:
Say if you have two tmux window, one is vim, the vim has buffer with
```
ls
ls
```
The other one is opened shell (either you're on fish, zsh). 
1st case: Before this commit, set `g:slime_bracketed_paste = 1`,  in vim window:
```
SlimeSend1 ls
```
will send `ls\r` to another shell, but the `ls` will not got executed.

2nd case: in vim window, either (select all the content, `<Plug>SlimeRegionSend`) or `<Plug>SlimeParagraphSend`, will both send `ls\n\ls\n` to another shell, but the two `ls` command will not be executed.

After the commit. All the commands will be executed

jpalardy:
Looking good. I tried on my side, with and without bracketed_mode.

Also, I see now how anything you could send inside the bracketed paste won't send the final `Enter`. (it's by design)

jpalardy:
Thanks for the PR ^

PR #332, opened 2022-06-13
Automatically create missing directories and detect errors when writing to `g:slime_paste_file`

danielwe:
I'm noticing that vim-slime will fail silently if the folders for `g:slime_paste_file` don't already exist, so here's a fix for that. (If you'd rather not have this, an alternative would be to detect the problem and emit a warning. It's just the current silence that's not ideal.)

The last commit also includes my opinionated change of the default `g:slime_paste_file` to `$HOME/.cache/vim-slime/paste`, but feel free to drop that. (EDIT: changed to use `$XDG_CACHE_HOME` with `$HOME/.cache` as fallback.)

### DIFF
--- a/autoload/slime.vim
+++ b/autoload/slime.vim
@@ -292,7 +292,10 @@ function! s:SID()
 endfun
 
 function! s:WritePasteFile(text)
-  call mkdir(fnamemodify(g:slime_paste_file, ":p:h"), "p")
+  let paste_dir = fnamemodify(g:slime_paste_file, ":p:h")
+  if !isdirectory(paste_dir)
+    call mkdir(paste_dir, "p")
+  endif
   let output = system("cat > " . g:slime_paste_file, a:text)
   if v:shell_error
     echoerr output


jpalardy:
Hi @danielwe 

Thanks for your PR ^

> an alternative would be to detect the problem and emit a warning

I would rather go with that approach. I feel that going down the rabbit hole of `$HOME/.whatever` vs `$HOME/.cache/dir/whatever` is a losing battle. (though it might not be in a few years, if adoption continues). It complicates the README and the code.

At some point, the "real" solution was to generate a tmp dir/file and to dump there. (or, even better, if tmux and others accepted STDIN, and there were no files involved at all)

But, I agree that an explicit error would be a lot better than a silent one.

jpalardy:
Hi @danielwe 

Thanks for your PR ^

> an alternative would be to detect the problem and emit a warning

I would rather go with that approach. I feel that going down the rabbit hole of `$HOME/.whatever` vs `$HOME/.cache/dir/whatever` is a losing battle. (though it might not be in a few years, if adoption continues). It complicates the README and the code.

At some point, the "real" solution was to generate a tmp dir/file and to dump there. (or, even better, if tmux and others accepted STDIN, and there were no files involved at all)

But, I agree that an explicit error would be a lot better than a silent one.

danielwe:
I hear you, there are very good reasons for a plugin like this keeping it simple and not bothering with XDG specs. Just thought I'd include it here as a suggestion and to show a use case for the first commit that adds the `mkdir` call. I'll roll that back.

But what you're saying is you'd also rather not have the `mkdir`? I can update the PR to detect and emit a warning instead, but I actually think that code would be more complicated than the single `mkdir` line.

autoload/slime.vim
```diff
@@ -292,7 +294,7 @@ function! s:SID()
 endfun
 
 function! s:WritePasteFile(text)
-  " could check exists("*writefile")
+  call mkdir(fnamemodify(g:slime_paste_file, ":p:h"), "p")
```
danielwe, line 297: To clarify, my core suggestion is just to add this line, which shouldn't require any changes to the readme or docs. (Forget about all the other stuff changing the default paste file.)

The alternative would be to check `v:shell_error` after the system call to detect whether it errored, and if so emit a warning, ideally including the output of the call in the warning.

Come to think of it, it may be good to do that anyway to catch other failure modes such as insufficient permissions.

jpalardy: sounds good: `mkdir` and create, with `v:shell_error` (catching most errors) 👍

danielwe:
OK, with the latest push:

* No defaults or docs are changed
* With `let g:slime_paste_file = expand("$HOME/.cache/vim-slime/paste")`, everything works even if the directory didn't exist before
* With `let g:slime_paste_file = "/slime_paste"`, a tmux send errors with the following message: Screenshot from 2022-06-14 22-07-54

autoload/slime.vim
jpalardy: is doing an `mkdir` on each call expensive?

would checking for the existence of the path enough? 🤔

danielwe: I was wondering about the same thing. My hunch was that, since `mkdir(..., "p")` is supposed to be a no-op when the directory already exists, it probably does its own check internally and returns early, so adding an explicit check would be redundant.

But who cares about hunches? Let's look at the code:

* `mkdir(..., "p")` checks and returns early here: https://github.com/vim/vim/blob/a34b4460c2843c67a35a2d236b01e6cb9bc38734/src/filepath.c#L1505. There is a bit of boilerplate before getting there, mostly normalizing the path and checking which arguments were given, but at least it never enters the recursion or actually tries to create any directories.
* `isdirectory(...)` just performs the check with minimal boilerplate: https://github.com/vim/vim/blob/a34b4460c2843c67a35a2d236b01e6cb9bc38734/src/filepath.c#L1417.

So I guess an explicit check is slightly more economical on the C side of things for the common case where the directory already exists. I have no idea whether the difference would be measurable in light of vimscript function call and branching overheads, and I won't bother profiling, but I'll go ahead and make the change---at least the code will look much less alarming.

PR #333, opened 2022-06-17
Wezterm support

uncomfyhalomacro:


autoload/slime.vim
```diff
@@ -63,6 +63,28 @@ function! s:KittyConfig() abort
...
+"""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""
+function! s:WeztermSend(config, text)
+	call system("wezterm cli send-text --no-paste --pane-id " . shellescape(a:config["pane_id"]) . " " . shellescape(a:text))
+endfunction
+
+function! s:WeztermConfig() abort
+	if !exists("b:slime_config")
+		let b:slime_config = {"pane_id": 0}
+	end
+	let b:slime_config["pane_id"] = str2nr(system("wezterm cli list --format json | jq '.[][\"pane_id\"]' | tail -n 1"))
```
wez, line 77: I'm not sure about this: wezterm doesn't guarantee an ordering for the data returned here, and the results might be surprising: for example, if you start with a tab with 2 panes for slime/repl, then create a new tab, this logic will probably select the pane from the new tab which is presumably not the desired outcome.  Connecting or detaching multiplexer domains may have other surprising consequences for the ordering of this data.

It's probably a good idea to take the current pane id from the `$WEZTERM_PANE` environment, then use `wezterm cli list --format json` to discover the tab_id for that pane, then select only the panes from that tab.

If `$WEZTERM_PANE` isn't set, then you could use logic similar to the second bullet in https://wezfurlong.org/wezterm/cli/cli/index.html#targeting-panes to find the least-idle session and use its focused pane id instead.

autoload/slime.vim
```diff
@@ -63,6 +63,28 @@ function! s:KittyConfig() abort
   let b:slime_config["listen_on"] = input("kitty listen on: ", b:slime_config["listen_on"])
 endfunction
 
+"""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""
+" Wezterm
+"""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""
+function! s:WeztermSend(config, text)
+	call system("wezterm cli send-text --no-paste --pane-id " . shellescape(a:config["pane_id"]) . " " . shellescape(a:text))
```
wez, line 70: I don't know anything about slime other than what I quickly skimmed from its readme, but it suggests that `g:slime_bracketed_paste` should control whether `--no-paste` is passed here

uncomfyhalomacro:
Thanks, i will update it later today.

jpalardy:
Hi @uncomfyhalomacro and @wez 

Thanks for dropping by 😄 

While I technically don't have anything against [wezterm](https://github.com/wez/wezterm), I'm getting worried by the increase of "targets" that slime supports… The combinatorial explosion of options and targets (in many combinations I don't personally use) could lead to some unplanned breakage 🤔 

While it might not be _SPECIFICALLY_ relevant to this PR, I've been wondering about a better way — probably with plugins — to support different targets. In practice, the "core" slime code has bindings and the logic to grab some text — and plugins receive a chunk of text and decide what they want to do with it.

The simplest contract would be to feed text to a command (user-definable) through STDIN (which would do away with shell escaping and other size limits)

Less clear is how to configure the plugin in a generic way.

---

If this is something you're interested in, we can iterate on this — with wezterm being the first experiment.

If not, we can discuss the merit of this plugin as per @wez's comments.

Let me know

jpalardy:
They are both [defined here](https://github.com/jpalardy/vim-slime/blob/main/plugin/slime.vim#L12-L13).

uncomfyhalomacro:
your suggestion is somewhat similar to https://github.com/hrsh7th/nvim-cmp where it will be just a completion framework so other people create their own plugins that can feed other sources of completions without the need to overcomplicate the "core" nvim-cmp.

so in summary, you want the "core" slime code to be a small plugin framework for other slime plugins to use so that any user-defined source e.g. kitty, wezterm, tmux, can be added without the need to modify the "core". seems like a good idea. i must say i have less experience with vim scripting probably because i'm one of the newer generations (i use neovim). i am willing to learn though 😄

jpalardy:
My own vimscript isn't amazing … I can think of 10 other languages I would rather use 😬 

But the focus is on:
- keeping things simple (code-wise)
- outsourcing the hard work (or, differently, plugins can be as complicated as they want)

So, to the extent where you were able to put this PR together, I think that's more than enough for this task.

jpalardy:
It's a new day: https://github.com/jpalardy/vim-slime-ext-plugins

- I removed almost everything (especially ftplugin)
- I'll push back (soon) a small core
- I'll personally iterate on tmux as my proof of concept
- feel free to watch there, and see how we can make this work

autoload/slime.vim
```diff
@@ -25,7 +25,8 @@ function! s:ScreenSend(config, text)
         \ " -X eval \"readreg p " . g:slime_paste_file . "\"")
   call system("screen -S " . shellescape(a:config["sessionname"]) . " -p " . shellescape(a:config["windowname"]) .
         \ " -X paste p")
-  call system('screen -X colon ""')
+  call system('screen -X colon "
```
jpalardy, line 28: any reason this is changed?

uncomfyhalomacro: oh probably i accidentally added a space. i closed the PR anyway so dont worry about it.

PR #336, opened 2022-07-07
Implement buffer-local slime_vimterminal_cmd setting

cwfoo:
Allow users to specify the default vimterminal command as a buffer-local variable. This would allow the default command to vary based on file type. For example:

```vim
autocmd FileType lisp let b:slime_vimterminal_cmd='sbcl'
autocmd FileType python let b:slime_vimterminal_cmd='python3'
```


jpalardy (approved):
Looking good, thanks ^

PR #341, opened 2022-08-23
use julia's begin end as if using ipython's %cpaste

junyixu:


jpalardy:
hi @junyixu 

I looked at the code: does this always work? does this need a conditional? (all versions of julia? … that type of thing)

ftplugin/julia/slime.vim
```diff
@@ -0,0 +1,8 @@
+function! _EscapeText_julia(text)
+	if len(split(a:text,"\n")) > 1
+		return ["begin\n", a:text, "end\n"]
+	else
+		return a:text
+	endif
+endfunction
```
jpalardy, line 7:
```suggestion
function! _EscapeText_julia(text)
  if match(a:text, "\n.") > -1
    return ["begin\n", a:text, "end\n"]
  end
  return a:text
endfunction
```

- changed if to only look for newline
- changed indent (2 spaces)
- no need for `else` on `if-return`

junyixu:
Thanks for your editing!

It works at least above julia version 1 as far as I know. (i.e. all versions of julia. I don't think one has a reason to use the version below v1)

If codes are surrounded by "begin ... end", only the result of the last line of the code will be shown and `;` doesn't work.
But one can use `println(val)` in the middle. 

Maybe something like `g:slime_python_ipython` can be added? I don't know...

jpalardy:
please update the PR and I'll merge it 👍

jpalardy:
Thanks again ^

I might cc you on issues from other Julia users, if it comes up.

fredrikekre:
What is the reason for this? If there is a good reason, how do I disable it? This now clutters the history with `begin`/`end`s.

junyixu:
@jpalardy 

- `match(a:text, "\n") > -1` doesn't work and `len(split(a:text,"\n")) > 1` works. Now one line is also surrounded by "begin ... end". I don't know why.
image


- this is better: 
  1. delete empty lines 
  2. exclude codes already surrounded by "begin" and "let"

```vim
function! _EscapeText_julia(text)
	if len(split(a:text,"\n")) > 1 
		let empty_lines_pat = '\(^\|\n\)\zs\(\s*\n\+\)\+'
		let no_empty_lines = substitute(a:text, empty_lines_pat, "", "g")
		if no_empty_lines[0:2]=="let" || no_empty_lines[0:4]=="begin"
			return no_empty_lines
		else
			return ["begin\n", no_empty_lines, "end\n"]
		endif
	endif
	return a:text
endfunction
```

fredrikekre:
> It's not recommended that multiple expressions in one cell.

That is specific to Pluto.jl notebooks, which is enforces a *very* opinionated style, and not a general Julia recommendation.

Personally I don't like that the code to be executed is modified like this. I really liked that sending code to the REPL was more or less equivalent to typing it myself. IMO this should be enabled by default.

> delete empty lines

I don't think you can do this without a Julia parser. You don't know if the empty line matters or not (could be inside a string, for example).

junyixu:
@fredrikekre 
> could be inside a string, for example

I'm afraid I don't understand the use case:
```julia
str="\n\n"
```
image

junyixu:
@fredrikekre 
Try it
```vim
function! _EscapeText_julia(text)
	if  exists('g:slime_julia') && len(split(a:text,"\n")) > 1 
		let empty_lines_pat = '\(^\|\n\)\zs\(\s*\n\+\)\+'
		let no_empty_lines = substitute(a:text, empty_lines_pat, "", "g")
		if no_empty_lines[0:2]=="let" || no_empty_lines[0:4]=="begin"
			return no_empty_lines
		else
			return ["begin\n", no_empty_lines, "end\n"]
		endif
	endif
	return a:text
endfunction
```
I bet you will enjoy it especially when you are plotting a figure with some maps, for exmaple

```vim
nmap <silent><buffer> <CR> :exec "normal \<Plug>SlimeLineSend"<cr>j
nnoremap <silent><buffer> <localleader>r :exec "normal \<Plug>SlimeSendCell"<cr>
nmap <silent><buffer> <space><space> <Plug>SlimeParagraphSend
```
you can use `<CR>` send a line and go to next line.

Klafyvel:
I strongly agree with @fredrikekre. Enclosing code within `begin`/`end` should not be the default behaviour of vim-slime with Julia. You are referring to very specific use cases using (I guess ?) Plots.jl, but most of the time you want to evaluate each line as if you typed it.

fredrikekre:
> I'm afraid I don't understand the use case:

I don't understand your example and what it is supposed to show. But for example here:
```
str = """
line 1

line 2
"""
```
you can not simply drop the empty line.

> I bet you will enjoy it [...]

Well, I am only complaining here because it *does* annoy me and I *don't* enjoy it :)

junyixu:
@fredrikekre 
Have you really tried the code?
```vim
function! _EscapeText_julia(text)
	if  exists('g:slime_julia') && len(split(a:text,"\n")) > 1 
		let empty_lines_pat = '\(^\|\n\)\zs\(\s*\n\+\)\+'
		let no_empty_lines = substitute(a:text, empty_lines_pat, "", "g")
		if no_empty_lines[0:2]=="let" || no_empty_lines[0:4]=="begin"
			return no_empty_lines
		else
			return ["begin\n", no_empty_lines, "end\n"]
		endif
	endif
	return a:text
endfunction
```
`exists('g:slime_julia')` will not change the default behaviour of vim-slime, just as what `g:slime_python_ipython` does.
Just try it.

Also,
@fredrikekre `str = """
line 1


line 2
"""`
then julia will throw an error.
what about deleting two empty lines?

junyixu:
@Klafyvel 
> You are referring to very specific use cases using (I guess ?) Plots.jl,

https://github.com/jpalardy/vim-slime/pull/341#issuecomment-1230514312

using PyPlot and it's not a very specific use cases.
I think many python users use `plt.subplots`

Klafyvel:
Well plotting using PyPlot.jl *is* a specific use case for Julia (there are even people who don't do plotting ;) ). Plus, the example you give works as expected without `begin`/`end`, could you elaborate on why this is not satisfactory?

junyixu:
😭 @Klafyvel the last para of https://github.com/jpalardy/vim-slime/pull/341#issuecomment-1230573161

junyixu:
the solution is to redraw it or use `begin/end`. I personally think Pluto.jl notebooks is reasonable.

Klafyvel:
This is not what happens when I run it, am I missing something? (I am not familiar with Plots.jl, sorry)

20220829_19h46m47s_grim




> the solution is to redraw it or use `begin/end`. I personally think Pluto.jl notebooks is reasonable.

Redrawing it sounds like the appropriate solution. Enforcing the use of `begin/end` blocks for everyone is a bit unreasonable. As @fredrikekre said, Pluto's choice is very opinionated and thus does not reflect the preference of the entire community. As a plugin for an editor, vim-slime should not adopt such a stance by making it the default behaviour. It can however be an optional one. The same goes for your proposal of stripping empty lines; as demonstrated previously, there *are* use cases where you want those blank lines, *e.g.* when you have a long markdown text in a multiline string. I don't really see a use case where you would like to remove those lines, but if you have one it should be an optional (preferably distinct from the one adding `begin/end` blocks) feature.

junyixu:
For python users familiar with python and matplotlib, he/she may like to use PyPlot.

Redrawing sounds like a bad solution. every time you plot, you need to resize the fig with your mouse or use `    fig.canvas.draw()` to assure redrawing. It's a pain in the ass. 

If you look at the source code of PyPlot, you find the julia codes above equal python's codes:
```python
fig, ax= plt.subplots(1, 2, figsize=(20, 10))
fig.show()
ax[1].plot(rand(3))
ax[2].plot(rand(5))
```
Because you didn't use `begin/end` julia thought you had done and executes `fig.show()`. So I said Pluto.jl notebooks' way to run a cell is reasonable.

I propose a way to run a cell with `begin/end` as if using python's `%cpaste`. and the last para of https://github.com/jpalardy/vim-slime/pull/341#issuecomment-1226777761
> Maybe something like g:slime_python_ipython can be added? I don't know...

For anyone don't know what I mean, I've given the code after that.

I'd like to see different opinions and you will absolutely solve your problem with the codes above?

> This now clutters the history with begin/ends.

But I wonder why history is important.

 what about deleting two continuous empty lines rather than one line since julia throws an error too?

junyixu:
I mean you can't type `str="line 1\n\n\nline 2"` in your way

fredrikekre:
>  Have you really tried the code?

I was commenting on the current state of master (278dd40).

> Also, @fredrikekre str="line 1\n\n\nline 2" is better. if you type `str = """
> line 1
> 
> line 2
> """`
> then julia will throw an error.

It is not better to write `str="line 1\n\n\nline 2"` if I want to write
```
str="""
line 1

line 2
"""
```
and there is no error:

```
julia> str = """
       line 1

       line 2


       line 3
       """
"line 1\n\nline 2\n\n\nline 3\n"

julia>
```

Multiline strings exist in the language for a reason. But this is not the point, my point is simply that you can not remove lines blindly as you suggest without knowing what they mean, and this require a Julia parser to properly handle.

But all of this is a moot discussion IMO, I just don't think vim-slime should be this opinionated when the language does not require it. This seems trivial enough that you can put it in your own RC file, alternatively include it here as something opt-in rather than forcing this onto everyone.

junyixu:
@fredrikekre did the code solve your problem?

Klafyvel:
> Because you didn't use `begin/end` julia thought you had done and executes `fig.show()`. So I said Pluto.jl notebooks' way to run a cell is reasonable.

This is not true? As I have shown before, matplotlib should have updated the plot as soon as you run `ax[2].plot(rand(3))`. 


> I'd like to see different opinions and you have absolutely solved your problem with the codes above, haven't you?

Making the behaviour you added optional indeed solves the issue, but that is not the current behaviour of the code that has been merged.

> But I wonder why history is important.

Because one might want to use the REPL in a classical way even if vim-slime is being used.

> I mean you can't type `str="line 1\n\n\nline 2"` in your way

I definitely can :
```julia
julia> str="line 1\n\n\nline 2"
"line 1\n\n\nline 2"
```

It seems to me that you are trying to replicate the behaviour of Pluto.jl in vim. You might be interested in using [this feature](https://github.com/fonsp/Pluto.jl/issues/240) rather than trying to copy Pluto's behaviour in vim-slime.

junyixu:
image

@fredrikekre if you type 3 `\n` s , julia will throw a error?

I mean, I can modify the code to delete repeat `\n`s

Klafyvel:
PyPlot.jl documentation explicitely states : 

> If you use PyPlot from an interactive Julia prompt, such as the Julia [command-line prompt](http://docs.julialang.org/en/latest/manual/interacting-with-julia/) or an IJulia notebook, then plots appear immediately after a plotting function (plot etc.) is evaluated.
Source : https://github.com/JuliaPy/PyPlot.jl#non-interactive-plotting

Is your REPL in some kind of specific state ? maybe a specific version or a specific `startup.jl` ? both behaviours you show in your two latest messages are unusual.

junyixu:
julia version 1.7.2

fredrikekre:
@fredrikekre if you type 3 \n s , julia will throw an error?

You can make sure to escape the direct eval there by meta-enter, regardless, this is valid syntax (try in a source file) and you should be able to use vim-slime with valid syntax. But again, this is beside the point here.

junyixu:
how vim-slime ` escape the direct eval there by meta-enter`

fredrikekre:
You might need "bracketed paste mode", I have:
```
:let g:slime_bracketed_paste=1
```
for example.

But again, this is beside the point. **vim-slime should not change the source code.**

junyixu:
have
```vim
function! _EscapeText_julia(text)
	if  exists('g:slime_julia') && len(split(a:text,"\n")) > 1 
		let empty_lines_pat = '\(^\|\n\)\zs\(\s*\n\+\)\+'
		let no_empty_lines = substitute(a:text, empty_lines_pat, "", "g")
		if no_empty_lines[0:2]=="let" || no_empty_lines[0:4]=="begin"
			return no_empty_lines
		else
			return ["begin\n", no_empty_lines, "end\n"]
		endif
	endif
	return a:text
endfunction
```
solved your problem, yes or no

fredrikekre:
My primary concern is fixed, since that is opt-in. However, as I have (tried to) explain multiple times, that still changes the source code to something else since it removes empty lines.

jpalardy:
based on the feedback in this thread, I'm ready to roll this back this change ^

can I get thumbs up/down to confirm?

junyixu:
I think an option could be added to help those people who prefer `begin/end` style
```vim
function! _EscapeText_julia(text)
	if exists('g:slime_julia') && len(split(a:text,"\n")) > 1 
		let empty_3_lines_pat = '\(^\|\n\)\zs\(\s*\n\+\)\{3,}'
		let no_empty_3_lines = substitute(a:text, empty_3_lines_pat, "\n\n", "g")
		if no_empty_3_lines[0:2]=="let" || no_empty_3_lines[0:4]=="begin"
			return no_empty_3_lines
		else
			return ["begin\n", no_empty_3_lines, "end\n"]
		endif
	endif
	return a:text
endfunction
```
does
1. provide an option `g:slime_julia`
2. delete 3 repeat empty lines

junyixu:
as https://github.com/jpalardy/vim-slime/issues/194#issue-425747898 mentioned

junyixu:
`begin/end`  does no harm except for history.

Nobody will notice the change if he/she doesn't set the option `g:slime_julia`

jpalardy:
hi @junyixu 

could you prepare a PR with these changes?

fredrikekre:
> begin/end does no harm except for history.

Well, that is highly subjective. It will make it more difficult to search, for example. And it does ruin history in the sense that it includes more things than what I evaluated.

> Regardless, nobody will notice the change if he/she doesn't set the option g:slime_julia. That's enough for some one who hold the strange point that vim-slime should not change the source code.

Why is that strange? Wouldn't you be annoyed if vim-slime changed your variable names, for example? Or removed some elements of an array you evaluated? Again, I don't understand why lines should be removed at all.

jpalardy:
Reverted for now: c959072

further work in #343

PR #342, opened 2022-08-30
use julia's begin end as if using ipython's %cpaste

junyixu:
1. provide an option g:slime_julia
2. only substitute `\{3,}` repeat empty lines

as https://github.com/jpalardy/vim-slime/pull/341#issuecomment-1231107918 mentioned


junyixu:
Sorry for the indent🤕

PR #343, opened 2022-08-30
use julia's begin end as if using ipython's %cpaste with modification

junyixu:
1. provide an option `g:slime_julia`
2. delete repeat empty lines

### DIFF
--- a/ftplugin/julia/slime.vim
+++ b/ftplugin/julia/slime.vim
@@ -1,7 +1,12 @@
 function! _EscapeText_julia(text)
-  if match(a:text, "\n") > -1
-    return ["begin\n", a:text, "end\n"]
-  end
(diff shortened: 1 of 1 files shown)


junyixu:
fixed https://github.com/jpalardy/vim-slime/pull/341

jpalardy:
Hi @junyixu 

I reverted main for now c959072

Right now, the logic in your PR is a bit _complicated_. Is there a way to simplify this?

Also, in the meantime, you can add your function to your vim config as:

```vim
function SlimeOverride_EscapeText_julia(text)
" ...
endfunction
```

Have a look at [language transformations](https://github.com/jpalardy/vim-slime#how-to-override-language-transformations), if that helps.

junyixu:
I've already shared my code for someone needing it to modify it himself.

I have no time now.

junyixu:
And thanks for your work

PR #346, opened 2022-10-07
Add handler for sh

graywolf:
If tabs are send into a bash, they trigger bash completion and that results in the REPL not really working as expected. So we translate tabs into spaces, which in context of shell is usually safe operation. There might be some edge cases, but it definitely fixes more things then it breaks.


jpalardy:
hi @graywolf 

looks good, thanks for pitching in!

PR #352, opened 2022-10-29
feat: allow user to specify a function to query the jobid

AckslD:


jpalardy:
Hi @AckslD 

That looks good to me, thanks for the PR 👍 

(if it causes problem, I might reach out)

AckslD:
Cool, thanks for a quick reply and yes feel free to reach out if it causes issues :)

PR #353, opened 2022-11-17
move to the end of selected Visual area

tanloong:
After SlimeRegionSending, the cursor stays at the start of the selected Visual area. Moving the cursor to the end side would be helpful. It will allow us to just continue to select and send lines that come after.

### DIFF
--- a/autoload/slime.vim
+++ b/autoload/slime.vim
@@ -357,7 +357,7 @@ function! slime#send_op(type, ...) abort
   let rt = getregtype('"')
 
   if a:0  " Invoked from Visual mode, use '< and '> marks.
-    silent exe "normal! `<" . a:type . '`>y'
+    silent exe "normal! `<" . a:type . "`>y" . "`>"
   elseif a:type == 'line'
     silent exe "normal! '[V']y"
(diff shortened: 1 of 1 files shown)


jpalardy:
hi @tanloong 

Thanks for pitching in!

Is it possible to do this in a mapping instead? Does it belong (as the default behavior) in the slime logic?

jpalardy:
Hi @tanloong 

Although it's not _exactly_ the same, I've used [this binding](https://github.com/jpalardy/dotfiles/blob/main/vim/plugins.vim#L28) to navigate after sending a paragraph.

Also, I just tried:

```
# mnemonic: end
vmap ,e <c-c><c-c>`>
```

which seems to work (if I understand what you were trying to do)

Let me know if that helps 😄

tanloong:
It works! I use this:
```
xmap <F5> <c-c><c-c>`>
```
lua equivalent:
```
vim.keymap.set('x', '<F5>', '<c-c><c-c>`>', {remap=true})
```
Thanks!

jpalardy:
Happy to hear that ^

Issue #329, opened 2022-06-07
ZelliJ support


Suyashtnt:
ZelliJ is a multiplexer that I and a lot of other people use. Adding support for it would be great
https://zellij.dev/

jpalardy:
Hi @Suyashtnt 

I'm not against it, but someone will have to write/maintain it.

Let me know if that's something you would consider 👍

PR #355, opened 2023-01-13
Add zellij support

albheim:
Zellij is a terminal multiplexer like tmux or screen, and I (as well as others #329) think it would be great if vim-slime could support it. 

I had a quick try at implementing the needed functionality for vim-slime to work by using the reasonably new `action` subcommand in zellij.
I have never really tried vim-script before other than very basic vimrc stuff, so the current implementation is mostly copy pasted from the other implementations in vim-slime glued together with some guessing.
It seems to mostly work with regards to the zellij interaction, though I had some weird characters popping up in the vim window when running under WSL that I don't get for tmux, but on my regular linux install it all works fine.

You can set the session name you want to send to, default is `current`, as well as the relative pane in relation to the currently active pane, default is `current`, other options are `up/down/left/right` since those are the keywords available for pane navigation.

I have not yet added anything to the readme or documentation, thought I would first check if this would be acceptable? And if more should be done, get an idea of exactly what is required.


jpalardy:
Hi @albheim 

Thanks for the PR ✅ 

It's looking good, a few questions:
- if it needs support, will you be willing to pitch in? (I would cc on issues)
- can you add some documentation in both `README.md` and `doc/vim-slime.txt`? (it doesn't have to be long, check what other targets have done)

albheim:
I wouldn't mind having a look if you cc.

Added some docs now, do you have a preference where in the order they are added? Currently I just squeezed them in after kitty since that is where I put the implementation.

autoload/slime.vim
```diff
@@ -25,7 +25,8 @@ function! s:ScreenSend(config, text)
         \ " -X eval \"readreg p " . g:slime_paste_file . "\"")
   call system("screen -S " . shellescape(a:config["sessionname"]) . " -p " . shellescape(a:config["windowname"]) .
         \ " -X paste p")
-  call system('screen -X colon ""')
+  call system('screen -X colon "
+"')
```
jpalardy, line 29: is that a merge/reformat include?

let's revert it?

jpalardy (commented):
looking good 🚀 

I left a small comment to revert the invisible newline 😄

jpalardy:
> Currently I just squeezed them in after kitty since that is where I put the implementation.

that's fine — I'm especially grateful for the documentation patch, I won't be picky about the where 👍

autoload/slime.vim
albheim: Ahh, yeah. Had the same problem for the initial commit, but then I noticed it. Forgot about it for this small fix.

jpalardy:
Thanks again 😄

PR #366, opened 2023-02-25
Fix creating emtpy directory when not resolving environment variables

### LINKED ISSUE #365: Bug: vim-slime creates an empty directory when using shell environment variables in config file
First of all, thanks for all your work on vim-slime, it is one of my favorite vim plugins!
  
#### Description:
I have noticed a subtle bug: When specifying an alternative `g:slime_paste_file`  in the vimrc file which contains a shell environment variable, e.g. `$HOME`, vim-slime does not resolve it and creates a directory called `$HOME` in the directory from which vim currently operates.

#### Desired behavior:
The directory should not be created, i.e. the variable should be resolved.

#### Steps to reproduce:
* Put this minimal example in your vimrc file:
```
let g:slime_paste_file = "$HOME/.other_file_name"
nmap <c-c><c-c> <Plug>SlimeParagraphSend
```
* Open a file and use `SlimeParagraphSend` to send something to the REPL
* There will be an empty directory `$HOME` in the current directory.

I will submit a pull request later which should resolve the issue.

maisevector:
Thanks for your work on vim-slime! This PR resolves creating an empty directory when specifying a paste-file with environment variable and therefore closes #365 

The directory is created in line 364. This only happens when the `g:slime_paste_file` is explicitly specified in the vimrc-file since the default one from vim-slime is resolved in line 15.

A simple `expand()` on line 362 makes vim expand `$HOME` or whatever environment file there is and resolves the issue for me.

### DIFF
--- a/autoload/slime.vim
+++ b/autoload/slime.vim
@@ -359,7 +359,7 @@ function! s:SID()
 endfun
 
 function! s:WritePasteFile(text)
-  let paste_dir = fnamemodify(g:slime_paste_file, ":p:h")
+  let paste_dir = fnamemodify(expand(g:slime_paste_file), ":p:h")
   if !isdirectory(paste_dir)
     call mkdir(paste_dir, "p")
   endif


FreedomBen:
Tested, confirmed working.  Fixes the bug!  Thanks @maisevector

FreedomBen:
If anyone else has empty directories called the literal '$HOME' littered all over their filesystem, this is how I cleaned it up:

```find . -type d -name '$HOME' -exec rmdir {} \;```

jpalardy:
Hi @maisevector and @FreedomBen 

As discussed on #365, I think the fix is in the documentation.

Reasoning:
- not expanding $variable in strings is how vim works
- though _VERY_ unlikely, as user _MIGHT_ want to use a `$` in their path, without it being auto-expanded

Issue #371, opened 2023-03-14
Question with vimterminal


jiz4oh:
Hi @jpalardy @gbarta, I am curious which problem is metioned [here](https://github.com/jpalardy/vim-slime/blob/bea75d10247cc81632747084c290f4a7138e7d69/autoload/slime.vim#L265)

I have the question because I want to send below codes to my REPL

```ruby
has_many_option_values = option_values_variants
  .where(option_value: product_option_type.option_type.option_values.unscope(:order))
  .unscope(:order)
  .group(:option_value_id)
  .having("count(spree_option_values_variants.variant_id) > 1").pluck(:option_value_id)
```

but it's mess up my console

<img width="1043" alt="image" src="https://user-images.githubusercontent.com/41264693/225070261-e9775f48-d77b-4a48-982b-d8f490b2ed56.png">

extra codes are displayed, but what I want to display is like the one in the following figure

<img width="1045" alt="image" src="https://user-images.githubusercontent.com/41264693/225070659-a16bce14-1aee-4a5c-afe9-22ebbb3e25d7.png">

I just update this [codes](https://github.com/jpalardy/vim-slime/blob/bea75d10247cc81632747084c290f4a7138e7d69/autoload/slime.vim#L268)

<img width="620" alt="image" src="https://user-images.githubusercontent.com/41264693/225072403-91b5ee32-7a16-4baf-8236-f2df0a78435f.png">

could we instead use a single term_sendkeys call as metioned in the commen now?

references: https://github.com/jpalardy/vim-slime/commit/8f6533946bfaa8d08d7fb42a1a195babe607682e

gbarta:
The workaround no longer seems to be required.

jiz4oh:
thanks a lot @gbarta  @jpalardy

PR #372, opened 2023-03-14
Remove obsolete workaround for vimterminal problem

gbarta:
Fix for #371

Workaround was put in place while the vimterminal feature was new and still being developed.  It no longer appears to be required.


jpalardy:
looking good 👍 

I don't use vimterminal myself, so we'll see if it breaks for other people.

Issue #373, opened 2023-04-02
Issue using with wezterm+ipython


MadFisa:
Hi,

I have been using vim-slime with tmux+ipython for some time and is quite happy with it. Thank you for this fantastic plugin.

However, recently I have been trying out wezterm and was very happy to see that vim-slime supports wezterm. I tried it out, but is facing some issues.  When I use SlimeSend it copies the line to ipython terminal, but does not execute, instead adding a new line at the end. This is basically same as #208 , except for wezterm.  I haven't tried the fix yet since I am not comfortable with vim script. 
Similarly, if I select multiple lines and try SlimeSend it copies the line, but does not execute and I get a '--' at the end. This is also similar to this issue #299 I found.

I tried using the exact same configuration with tmux and it worked as expected without any problems. My guess is that there is some problem with the handling of newline character '\n'. I don't know how to further troubleshoot this. Any suggestions would be very appreciated.

I am using neovim if that would make any difference.

jpalardy:
@albheim @Strosel thoughts?

MadFisa:
@jpalardy sorry if this is a basic question, I do not know vimscript.
What are these lines supposed to do?

https://github.com/jpalardy/vim-slime/blob/e5ae6a6d479c97d5352969574130956798e480b7/autoload/slime.vim#L165-L169

MadFisa:
@albheim  Thanks for looking into it and the explanation. Currently, with my limited understanding, I think there are two possibilities since this exact same configuration works with tmux:

1.  The issue is due to vim-slime having something in tmux codes that take care of the new line character that is not being done in the wezterm parts.
2. The newline character is handled differently by the wezterm compared to tmux

Can you point me towards where the relevant code for wezterm is?

albheim:
Here is wezterm, and right below is tmux. It might very well be possible to take some of the newline specific things from tmux to get wezterm to work, but I don't really know why and how this is handled differently so I can't really give any good tips.

https://github.com/jpalardy/vim-slime/blob/e5ae6a6d479c97d5352969574130956798e480b7/autoload/slime.vim#L105-L117

MadFisa:
@albheim  I think I have some clue to what is happening. Assuming what this part of the code does is basically pipe the selected text through `wezterm cli send-text --pane-id=`, I did some experiments with the command.

`echo 'some-text-here' |wezterm cli send-text --pane-id=2 ` will automatically insert a new line after the text.

[...]

PR #374, opened 2023-04-04
Adding bracketed paste option for wezterm

MadFisa:
A fix for issue #373. Implements option to turn on and off bracketed paste for wezterm.

### DIFF
--- a/autoload/slime.vim
+++ b/autoload/slime.vim
@@ -114,9 +114,9 @@ function! s:WeztermSend(config, text)
     let bracketed_paste = 0
   endif
   if bracketed_paste
-    call system("echo " . shellescape(a:text) . " | wezterm cli send-text --pane-id=" . shellescape(a:config["pane_id"]))
+      call system("wezterm cli send-text --pane-id=" . shellescape(a:config["pane_id"]). ' '. shellescape(a:text) )
   else
-    call system("echo " . shellescape(a:text) . " | wezterm cli send-text --no-paste --pane-id=" . shellescape(a:config["pane_id"]))
+      call system("wezterm cli send-text --no-paste --pane-id=" . shellescape(a:config["pane_id"]). ' '. shellescape(a:text))
   endif
   endif
 endfunction
--- a/autoload/slime.vim
+++ b/autoload/slime.vim
@@ -25,7 +25,8 @@ function! s:ScreenSend(config, text)
         \ " -X eval \"readreg p " . g:slime_paste_file . "\"")
   call system("screen -S " . shellescape(a:config["sessionname"]) . " -p " . shellescape(a:config["windowname"]) .
         \ " -X paste p")
-  call system('screen -X colon ""')
+  call system('screen -X colon "
+"')
 endfunction
 
 function! s:ScreenSessionNames(A,L,P)
@@ -113,11 +114,11 @@ function! s:WeztermSend(config, text)
   else
     let bracketed_paste = 0
   endif
-
   if bracketed_paste
-      call system("echo " . shellescape(a:text) . " | wezterm cli send-text --pane-id=" . shellescape(a:config["pane_id"]))
+    call system("echo " . shellescape(a:text) . " | wezterm cli send-text --pane-id=" . shellescape(a:config["pane_id"]))
   else
-      call system("echo " . shellescape(a:text) . " | wezterm cli send-text --no-paste --pane-id=" . shellescape(a:config["pane_id"]))
+    call system("echo " . shellescape(a:text) . " | wezterm cli send-text --no-paste --pane-id=" . shellescape(a:config["pane_id"]))
+  endif
   endif
 endfunction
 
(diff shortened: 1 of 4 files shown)


albheim:
Looks reasonable, and "works" for the cases I tested. Though the double enter annoys me.

Especially since it seems like it does not do it from only wezterm, i.e. 
```
echo ls | wezterm cli  send-text --pane-id=18 --no-paste
```
correctly runs `ls` in pane 18 for me, and it must be the vim-slime that adds an extra newline for some reason.

Might be less disruptive for people who use this to have bracketed paste off by default since that is how it was before? Though maybe it is nicer to align with tmux settings?

albheim:
What about trying to work in this bit also? Seems like it would deal with exactly removing the trailing stuff.
https://github.com/jpalardy/vim-slime/blob/e5ae6a6d479c97d5352969574130956798e480b7/autoload/slime.vim#L142-L149

Though it would actually need to run in both cases, not only in bracketed mode, if we want to get rid of the double enter we see.

autoload/slime.vim
```diff
@@ -106,7 +106,19 @@ endfunction
...
+    let bracketed_paste = g:slime_bracketed_paste
+  else
+    let bracketed_paste = 0
+  endif
+
+  if bracketed_paste
+      call system("echo " . shellescape(a:text) . " | wezterm cli send-text --pane-id=" . shellescape(a:config["pane_id"]))
+  else
+      call system("echo " . shellescape(a:text) . " | wezterm cli send-text --no-paste --pane-id=" . shellescape(a:config["pane_id"]))
+  endif
```
albheim, line 121:
```suggestion
  if bracketed_paste
    call system("echo " . shellescape(a:text) . " | wezterm cli send-text --pane-id=" . shellescape(a:config["pane_id"]))
  else
    call system("echo " . shellescape(a:text) . " | wezterm cli send-text --no-paste --pane-id=" . shellescape(a:config["pane_id"]))
  endif
```

autoload/slime.vim
```diff
@@ -25,7 +25,8 @@ function! s:ScreenSend(config, text)
         \ " -X eval \"readreg p " . g:slime_paste_file . "\"")
   call system("screen -S " . shellescape(a:config["sessionname"]) . " -p " . shellescape(a:config["windowname"]) .
         \ " -X paste p")
-  call system('screen -X colon ""')
+  call system('screen -X colon "
+"')
```
albheim, line 29: Something seems to have happened here, think I had the same thing with my editor wanting to change this.

MadFisa: Thanks for catching it. Don't know what happened there.

jpalardy:
Hi there

I’m following this, but I’m away from the computer for the week

I’ll reply properly next weekend

MadFisa:
> What about trying to work in this bit also? Seems like it would deal with exactly removing the trailing stuff.
> 
> https://github.com/jpalardy/vim-slime/blob/e5ae6a6d479c97d5352969574130956798e480b7/autoload/slime.vim#L142-L149
> 
> Though it would actually need to run in both cases, not only in bracketed mode, if we want to get rid of the double enter we see.

I tried it. Specifically the part that strips out the trailing endline character. It works but introduces a new problem with slime_python_ipython option, where it causes ipython console to freeze up while sending '--'. Not sure what is happening there.

MadFisa:
This reminds me of what happened here [here](https://github.com/jpalardy/vim-slime/issues/373#issuecomment-1495957049). Maybe there is something that happens with '--' (or `g:slime_python_ipython` option) and endline combination?

MadFisa:
@albheim @jpalardy 
So I figured out the problem and kinda have a solution. All the methods that removes that extra endline ( removing echo, removing using the tmux code snippet) causes an issue that when using `g:slime_python_ipython` when sending multiple lines. For some reason, '--' is not send to ipython console, if there is not a (extra?) endline before '--'. I tried and added an extra endline ftplugin/python before the '--' and it seems to have fixed the problem. I tested with wezterm and limited testting with tmux haven't found any problems so far. 
But since this requires modifying code that is not just specific to wezterm, I will leave this to discretion on whether to merge or not.

autoload/slime.vim
MadFisa: Reverted in the next commit

ftplugin/python/slime.vim
```diff
@@ -5,7 +5,7 @@ end
 
 function! _EscapeText_python(text)
   if exists('g:slime_python_ipython') && len(split(a:text,"\n")) > 1
-    return ["%cpaste -q\n", g:slime_dispatch_ipython_pause, a:text, "--\n"]
+    return ["%cpaste -q\n", g:slime_dispatch_ipython_pause, a:text, "\n--\n"]
```
jpalardy, line 8: I'm really hesitant about changing the python ftplugin here — especially if this only came up with wezterm

jpalardy:
Hi @MadFisa @albheim 

I downloaded wezterm and I'm going to experiment a bit.

Let me get back to you with my findings.

jpalardy:
@MadFisa @albheim 

I took a shot at it with #375, please have a look

If that works for you, I'll integrate the documentation from this PR into it

I tried this both with `g:slime_bracketed_paste = 1` and `g:slime_bracketed_paste = 0` (as per tmux's reference)

PR #375, opened 2023-04-10
Fix wezterm

jpalardy:
- use STDIN
- support bracketed and non-bracketed modes
- cascade global config correctly

diff is easier to follow, one commit at a time

### DIFF
--- a/autoload/slime.vim
+++ b/autoload/slime.vim
@@ -107,7 +107,7 @@ endfunction
 """""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""
 
 function! s:WeztermSend(config, text)
-  call system("echo " . shellescape(a:text) . " | wezterm cli send-text --pane-id=" . shellescape(a:config["pane_id"]))
+  call system("wezterm cli send-text --pane-id=" . shellescape(a:config["pane_id"]), a:text)
 endfunction
 
 function! s:WeztermConfig() abort
--- a/autoload/slime.vim
+++ b/autoload/slime.vim
@@ -58,7 +58,7 @@ function! s:KittyConfig() abort
   end
   let b:slime_config["window_id"] = str2nr(system("kitty @ select-window --self"))
   if v:shell_error
-    let b:slime_config["window_id"] = input("kitty window_id: ","1") 
+    let b:slime_config["window_id"] = input("kitty window_id: ","1")
   end
   let b:slime_config["listen_on"] = input("kitty listen on: ", b:slime_config["listen_on"])
 endfunction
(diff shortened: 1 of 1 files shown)


albheim:
Works fine except for python/ipython where if you send multiple lines it will either indent them incorrectly (if brackets paste is off) or it will require a manual enter after inserting them (if bracket paste is on).

Seems to be since python requires two newlines in a row for multiline things to mark that input is done.

MadFisa:
@jpalardy I am away from my pc for few days, I will take a look as soon as I get back.

jpalardy:
@albheim have you tried this with tmux? I wonder how it behaves 🤔

albheim:
It's the same for tmux.

Not saying it is something that has to be fixed, just reflected on it since it was part of the discussion in #373 with regards to double enters. 
I can't really think of any good way to fix it since it would probably require different handling of different targets, and I don't think you have any way of seeing what the target is (at least currently)?

I'm also questioning if having bracketed paste off by default is the right choice? I agree it should be consistent, and tmux already do it this way, though I don't really see many cases when I would prefer it that way. Manually pasting some code in e.g. ipython or julia repl seems to use bracketed paste for me (don't really know anything about this, but tried copying from some different sources so it wasn't that which made it), and staying consistent with "the normal experience" is also worth regarding.

jpalardy:
Hi @albheim 

Good point about `bracketed_paste` — it seems like a reasonable default now, but it certainly wasn't in play when vim-slime first came around.

I think the "fix" for this will be to add a strongly worded comment in the wezterm section of the documentation

MadFisa:
@jpalardy 
It works as expected as with `bracketed_paste` on (does not require manual enter for me ) with wezterm and tmux with both ipython mode on and off. Interestingly I do not require the manual enter as observed by @albheim .

Switching off bracketed paste creates extra indents when sending multiple lines which have indents in them. But I think that is the expected behavior with `bracketed_paste `off.  

So overall this seems to work as expected for me. Also I agree with @albheim about `bracketed_paste` being enabled should be the sensible default now.

albheim:
You tried with both single and multiline entries? 

Single worked fine for me (no extra enter needed), but for multiline it was needed. And this seems to be how the python repl works normally as well, since for single line you just enter once to run, but for multiline the first enter is just new line while second runs the entry.

MadFisa:
@albheim It is working for me without any problems as long as `bracketed_paste` is on (irrespective of ipython mode).
 Just to clarify, lines are sent to Wezterm by visually selecting them in neovim and then using `SlimeSend` to send them to the terminal.

albheim:
Hmm, interesting. 
I use vim so maybe a vim/neovim difference? Though that also seems a bit odd.

Well, either way I think this PR is an improvement so if anyone has problems with this in the future it can be looked at then.

PR #385, opened 2023-07-19
Send cancel command to tmux panel before pasting

randomizedthinking:
It is useful when the tmux panel is scrolled and in the browser mode. Or bracket-paste would not be in effect, and a manual return is required to run the pasted command.

### DIFF
--- a/autoload/slime.vim
+++ b/autoload/slime.vim
@@ -181,6 +181,7 @@ function! s:TmuxSend(config, text)
     let chunk = text_to_paste[i * chunk_size : (i + 1) * chunk_size - 1]
     call s:WritePasteFile(chunk)
     call s:TmuxCommand(a:config, "load-buffer " . g:slime_paste_file)
+    call s:TmuxCommand(a:config, "send-keys -X -t " . shellescape(a:config["target_pane"]) . " cancel")
     if bracketed_paste
       call s:TmuxCommand(a:config, "paste-buffer -d -p -t " . shellescape(a:config["target_pane"]))
     else


jpalardy:
hi @randomizedthinking 

so, this unconditionally sends a `cancel` to tmux — whether or not in copy mode?

randomizedthinking:
@jpalardy yes, it is to send a `cancel` unconditionally. It has no effect under the normal mode.

jpalardy:
let me experiment with this myself — I'll let you know 👍

autoload/slime.vim
```diff
@@ -182,6 +182,7 @@ function! s:TmuxSend(config, text)
     call s:WritePasteFile(chunk)
     call s:TmuxCommand(a:config, "load-buffer " . g:slime_paste_file)
     if bracketed_paste
+      call s:TmuxCommand(a:config, "send-keys -X -t " . shellescape(a:config["target_pane"]) . " cancel")
```
jpalardy, line 185: why is it inside the `if bracketed_paste` branch? 🤔

randomizedthinking: When `bracketed_paste` is off, the side effect of `cancel` is to bring focus to the bottom, but it does not affect paste. It is more a personal preference to send a `cancel` or not.

I intend not to mess with other people's workflow if they don't complain about it.

jpalardy: Can you update the PR to put the `cancel` just above the `if`?

I've experimented and it doesn't seem like a bad workflow.

If there's pushback, we can add a flag to make this conditional. (something like `g:slime_send_cancel` ?)

randomizedthinking: The requested update is done.

jpalardy:
Looking good, thank you 👍 

We'll give this a go and see if we need to change anything. That includes yourself, if you notice something.

PR #386, opened 2023-07-28
Fix the behavior of g:slime_python_ipython

Henderake:
The `g:slime_python_ipython` functionality is intended to be deactivated when its value is set to 0. However, an oversight in the _EscapeText_python function currently ignores this value, which results in the feature remaining enabled even when the flag is set to 0. This pull request rectifies this issue by incorporating the necessary check for this value.


jpalardy:
Hi @Henderake 

Good point, as it stands right now this is misleading 😬 

At worse, it doesn't hurt to be explicit.

Thanks for the patch 🎉

Issue #297, opened 2021-08-10
Not working with QtConsole


Freddyilha:
SO: ArchLinux

Hello, I have defined on my vimrc the 'let g:slime_target = "x11"'  and when I call *SlimeConfig* it asks which tmux socket and pane, although it works fine with tmux that's not what I want to use with QtConsole.
![image](https://user-images.githubusercontent.com/5214640/128795778-bc9fbc1d-b75d-4fb4-9c8d-b1e2fe67fb2d.png)

The *xdotool* is installed and working:
![image](https://user-images.githubusercontent.com/5214640/128795432-a300c7ea-34e6-
4a95-a4d0-f807498859c6.png)

Vim version and features:
![image](https://user-images.githubusercontent.com/5214640/128795032-14e1ae81-0570-4a0a-8a2e-711048b816e6.png)

Am I missing something?

jpalardy:
Hi @Freddyilha 

I just tried locally and it does fail silently (when it fails) 🤔 

@ynco any thoughts?

In the meantime, I would recommend peppering the plugin code with `echom` statements, and reading out the output in `:messages` (it's console.log for vim! -- [details](https://learnvimscriptthehardway.stevelosh.com/chapters/01.html))

Let me know what you find

Freddyilha:
Ok, so what I've discovered is that the variable is that the variable **slime_target** is really been ignored and it's stuck in tmux. I put a **echom** at each **section** and on each **function!** call to see what is been called, is there some kind of cache in vim?
![image](https://user-images.githubusercontent.com/5214640/128947881-53e4d855-744f-4aa9-92ac-f8063f369220.png)

jpalardy:
whoah…

although … this points towards your configuration not being picked up.

did you try to `echom` in your vimrc?  I wonder if something else is changing it… some things to try:
- temporarily disabling other plugins to remove "moving parts"
- check `scriptnames` to see what, and in what in order, is being loaded

Freddyilha:
Ok, I'm stupid, there was another definition of the slime_target at the end of my vimrc sorry for this.

But now I saw that it's only send the text without the Return command. Is it intended?

jpalardy:
hi @Freddyilha 

Happy to see you fixed some of it 😄 

> But now I saw that it's only send the text without the Return command. Is it intended?

that depends on the language plugin, and the target REPL ... what are the moving parts? and what's the behavior you see? (send me a screenshot, if possible)

[...]

PR #387, opened 2023-07-28
Add g:slime_x11_lf_to_cr flag

Henderake:
While using `vim-slime` with Jupyter QtConsole on Windows WSL, I experienced a multi-line issue similar to the one discussed in https://github.com/jpalardy/vim-slime/issues/297. Upon further investigation, it was discovered that these X11 applications do not break the lines on `\n` but rather on `\r`. This pull request introduces a new configuration variable `g:slime_x11_lf_to_cr`, enabling the substitution of `\n` with `\r` specifically when the target is X11.

### DIFF
--- a/autoload/slime.vim
+++ b/autoload/slime.vim
@@ -348,7 +348,14 @@ endfunction
 """""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""
 
 function! s:X11Send(config, text)
-  call system("xdotool type --delay 0 --window " . shellescape(b:slime_config["window_id"]) . " -- " . shellescape(a:text))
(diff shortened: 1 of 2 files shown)


Henderake:
Thanks for the guide. I've managed to use overrides in my .vimrc to solve it, so I'll close the PR.

```
let g:slime_x11_lf_to_cr = 1

function! SlimeOverride_EscapeText_python(text)
  let escaped_text = _EscapeText_python(a:text)
  if g:slime_target =~ "x11" && exists("g:slime_x11_lf_to_cr") && g:slime_x11_lf_to_cr
      return substitute(escaped_text, '\n', '\r', 'g')
  end
  return escaped_text
endfunction
```

PR #392, opened 2023-08-28
Add relative direction setting for wezterm

albheim:
Noticed that wezterm had the `wezterm cli get-pane-direction DIRECTION` option, meaning we can automatically get the pane id of a relative pane.

For now just added the option to set a default direction, since that is how I typically use it (i.e. I usually always have vim to the left and a repl to the right). 
But maybe it would be nice to allow for a relative direction in the prompt also instead of inserting the id in the case of a relative direction in the default setting.

For now I just wanted to push this since I added it locally for me to use.

### DIFF
--- a/autoload/slime.vim
+++ b/autoload/slime.vim
@@ -139,12 +139,8 @@ function! s:WeztermConfig() abort
   if !exists("b:slime_config")
     let b:slime_config = {"pane_id": 1}
   elseif exists("b:slime_config.pane_direction")
-    let tmp = system("wezterm cli get-pane-direction " . shellescape(b:slime_config["pane_direction"]), "\n")
-    if tmp != ""
-      let b:slime_config = {"pane_id": tmp}
-    else
-      let b:slime_config = {"pane_id": 1}
-    endif
+    let pane_id = system("wezterm cli get-pane-direction " . shellescape(b:slime_config["pane_direction"]))
+    let b:slime_config = {"pane_id": pane_id}
   endif
   let b:slime_config["pane_id"] = input("wezterm pane_id: ", b:slime_config["pane_id"])
 endfunction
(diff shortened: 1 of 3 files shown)


autoload/slime.vim
```diff
@@ -138,7 +138,14 @@ endfunction
 function! s:WeztermConfig() abort
   if !exists("b:slime_config")
     let b:slime_config = {"pane_id": 1}
-  end
+  elseif exists("b:slime_config.pane_direction")
+    let tmp = system("wezterm cli get-pane-direction " . shellescape(b:slime_config["pane_direction"]), "\n")
```
jpalardy, line 142: is the `\n` from STDIN necessary?

autoload/slime.vim
```diff
@@ -138,7 +138,14 @@ endfunction
...
+    if tmp != ""
+      let b:slime_config = {"pane_id": tmp}
+    else
+      let b:slime_config = {"pane_id": 1}
+    endif
+  endif
```
jpalardy, line 148: after line 140?

```vim
if exists("b:slime_config.pane_direction")
  let pane_id = system("wezterm cli get-pane-direction " . shellescape(b:slime_config["pane_direction"]))
  let b:slime_config = {"pane_id": pane_id}
endif
```

equivalent logic?

jpalardy (approved):
looking good — just a few comments

jpalardy:
hi @albheim 

Thanks for the PR 👍

autoload/slime.vim
albheim: My thought was if, e.g., we are in the rightmost pane and ask for the pane to the right, then we get an empty return and it would be nice to have a default. Though I guess this is just the suggestion for the user, so we can just leave it empty in that case. I'll update.

autoload/slime.vim
albheim: I'll check, you are probably correct. I just copied from another part of the file.

albheim:
Updated what you commented on, for me it can be merged.

jpalardy:
Looking good — @albheim, I might call on you if this code causes problem.

Thanks again

Issue #379, opened 2023-05-16
Bracketed paste not working with zellij


sourproton:
I have the following python file:

```
# %% cell 1
print("foo")

# %% cell 2
def bar(x):
    # compute y
    y = 2 * x

    return y

print(bar(1))

# %% cell 3
print("baz")
```

With `zellij` set up as the target with an `ipython` REPL and with `g:slime_bracketed_paste=1`, the `cell 2` bugs because of the blank line before the `return` statement. The same doesn't happen in tmux, where bracketed paste is working.

[According to](https://www.reddit.com/r/zellij/comments/13h8sgb/comment/jk7xyr9/?utm_source=share&utm_medium=web2x&context=3) Zellij's dev, for now a candidate fix would be to use `action write` instead of `action write-chars` and pass to it the binary equivalent of the bracketed string.

jpalardy:
Hi @sourproton 

`slime_bracketed_paste` was originally tmux-only ([here](https://github.com/jpalardy/vim-slime/blob/main/autoload/slime.vim#L170), and that's where it's documented in the README). Recently, it's been added to the wezterm target ([here](https://github.com/jpalardy/vim-slime/blob/main/autoload/slime.vim#L119))

Technically, there's no reason not to support it 😄 

If you want to have a stab at it, let me know.

sourproton:
Thank you for your answer!

As I understand, `vim-slime`s implementation of bracketed paste in `tmux` is done with the `-p` flag in `paste-buffer -d -p -t` [here](https://github.com/jpalardy/vim-slime/blob/bb152854fc21193729bf587aa055e738d50c8b48/autoload/slime.vim#LL186C53-L186C55), whilst `wezterm`'s implementation is done with the absence of the `--no-paste` flag [here](https://github.com/jpalardy/vim-slime/blob/bb152854fc21193729bf587aa055e738d50c8b48/autoload/slime.vim#LL128C24-L128C24).

So `vim-slime` doesn't actually parse and treat the string to be sent to the target (it does a bit with the trailing `\r` and `\n` with `wezterm` but it doesn't actually wrap the string with `\e[200~` and `\e[201~`), it uses the built-in cli command to achieve it.

Given that `zellij` doesn't feature this yet, and that might change, I don't see how this could be implemented right now, and I don't know if it's worth the effort to do it via `vim-slime`. I might take a look at `zellij`'s source and see if I can contribute to add a bracketed paste flag there and then use it here.

[...]

PR #394, opened 2023-09-23
bracketed paste for zellij

ctorney:
potential fix for #379 based on the suggestion from the zellij dev [here](https://www.reddit.com/r/zellij/comments/13h8sgb/how_to_send_text_to_split_pane/). This appears to work but maybe there's a better zellij-based solution on the way

### DIFF
--- a/autoload/slime.vim
+++ b/autoload/slime.vim
@@ -92,16 +92,12 @@ function! s:ZellijSend(config, text)
   endif
 
   if bracketed_paste
-    " zellij doesn't support bracketed paste so wrap in escape codes
-    let ec_start = [27, 91, 50, 48, 48, 126] " ESC [ 2 0 0 ~
-    let ec_end = [27, 91, 50, 48, 49, 126] " ESC [ 2 0 1 ~
+    call system("zellij " . target_session . " action write 27 91 50 48 48 126")
+    call system("zellij " . target_session . " action write-chars " . shellescape(text_to_paste))
+    call system("zellij " . target_session . " action write 27 91 50 48 49 126")
     if has_crlf
-      " trailing newline
-      let ec_end += [10]
-    end
-    " convert to list of ascii codes and stack the escape codes
-    let text_to_paste = ec_start + map(split(text_to_paste, '\zs'), 'char2nr(v:val)') + ec_end
-    call system("zellij " . target_session . " action write " . join(text_to_paste))
+      call system("zellij " . target_session . " action write 10")
+    endif
   else
     call system("zellij " . target_session . " action write-chars " . shellescape(text_to_paste))
   endif
--- a/autoload/slime.vim
+++ b/autoload/slime.vim
@@ -74,7 +74,38 @@ function! s:ZellijSend(config, text)
   if a:config["relative_pane"] != "current"
     call system("zellij " . target_session . " action move-focus " . shellescape(a:config["relative_pane"]))
   end
-  call system("zellij " . target_session . " action write-chars " . shellescape(a:text))
+  if exists("b:slime_bracketed_paste")
+    let bracketed_paste = b:slime_bracketed_paste
+  elseif exists("g:slime_bracketed_paste")
+    let bracketed_paste = g:slime_bracketed_paste
+  else
+    let bracketed_paste = 0
+  endif
+
+  let [text_to_paste, has_crlf] = [a:text, 0]
+  if bracketed_paste
+    if a:text[-2:] == "\r\n"
+      let [text_to_paste, has_crlf] = [a:text[:-3], 1]
+    elseif a:text[-1:] == "\r" || a:text[-1:] == "\n"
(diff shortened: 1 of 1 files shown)


autoload/slime.vim
```diff
@@ -74,7 +74,38 @@ function! s:ZellijSend(config, text)
...
+  if bracketed_paste
+    " zellij doesn't support bracketed paste so wrap in escape codes
+    let ec_start = [27, 91, 50, 48, 48, 126] " ESC [ 2 0 0 ~
+    let ec_end = [27, 91, 50, 48, 49, 126] " ESC [ 2 0 1 ~
+    if has_crlf
+      " trailing newline
+      let ec_end += [10]
+    end
+    " convert to list of ascii codes and stack the escape codes
+    let text_to_paste = ec_start + map(split(text_to_paste, '\zs'), 'char2nr(v:val)') + ec_end
```
jpalardy, line 103: if the goal is to generate a string with special characters, it might be simpler to use `\x##` inside strings, and prepend/append those strings

some example for elsewhere in the code:
- [scala ftplugin](https://github.com/jpalardy/vim-slime/blob/main/ftplugin/scala/slime.vim#L7)
- [coffescript ftplugin](https://github.com/jpalardy/vim-slime/blob/main/ftplugin/coffee/slime.vim#L4)

jpalardy:
Hi @ctorney 

Thanks for the PR 👍  — I commented on the string generation.

> maybe there's a better zellij-based solution on the way

Did you want to give this more time?

autoload/slime.vim
ctorney: you're right the escape sequences can go straight to strings - something like this?
```
 if bracketed_paste
    " convert to list of ascii codes and add the escape codes
    let text_to_paste = join(map(split(text_to_paste, '\zs'), 'char2nr(v:val)'))
    let ec_start = "27 91 50 48 48 126 " 
    if has_crlf
      " trailing newline
      let ec_end = " 27 91 50 48 49 126 10" 
    else
      let ec_end = " 27 91 50 48 49 126"  
    endif
    call system("zellij " . target_session . " action write " . ec_start . text_to_paste . ec_end)
  else
    call system("zellij " . target_session . " action write-chars " . shellescape(text_to_paste))
  endif
```
(it seems that the zellij write action only accepts a sequence of base-10 values)

ctorney:
Hi @jpalardy 
> Did you want to give this more time?

I can't see another solution in the pipeline looking at the zellij repo @sourproton - any thoughts on this?

sourproton:
Hi @ctorney

I like this solution a lot. I'm illiterate in vimscript and wasn't able to come to this solution myself.

Despite the solution being different from the other targets (i.e. for the others, their cli commands handle the bracket pasting, whereas for zellij, the string would be manipulated by vim-slime before being sent), I don't see it as a problem, and I too don't see this being addressed by zellij in the near future.

Even if zellij implements bracketed paste, I think it's worth it to continue with the PR now and merge it as an indefinite temporary solution. It will be very simple to modify vim-slime back to use zellij's cli bracketed paste flags once it exists.

autoload/slime.vim
jpalardy: I meant to go one step further and skip the join-map-split

something like:

```vim
let ec_start = "\x27[200~"
let ec_end = "\x27[201~"
if has_crlf
  let ec_end = ec_end . "\n"
endif
```

ctorney: the main issue is that the only way it appears to work correctly is by using the zellij action `write` rather than `write-chars` (write sends an array of bytes, whereas write-chars sends a string of characters).

so this works
```
zellij action write 27 91 50 48 48 126 104 101 108 108 111 10 104 101 108 108 111 10 27 91 50 48 49 126 10
```
but this doesn't
```
zellij action write-chars  "\x27[200~hello\nhello\x27[201~\n"
```
With your code we'd still have to include the join-map-split step to get the sequence of digits needed for the call to write. Maybe there's a way to do this using write-chars but I haven't been able to figure it out and I suspect it may be impossible because of the way zellij implements the two commands

jpalardy: damn 🤔 

as [it stands now](https://github.com/jpalardy/vim-slime/blob/main/autoload/slime.vim#L69-L81), it seems like the chars will be properly escaped for the shell.

and `write-chars` expects … a unicode string? and there's no loopholes for escape sequences?

meanwhile `write` probably hits some command-line argument limit relatively quickly 😬 

if there's no workaround, I think it's better to sandwich the code with the bracketedpaste sequences (with `if` no less), and keep the original `write-chars` for the real text?

ctorney: right - just updated. I thought I'd tried that before (write then write-chars then write again) and it didn't work but seems ok now so at least it gets rid of the join-map-split and the command line limit.

jpalardy:
looking good 🚀 

I think (short of something changing on the zellij's side) that this is a good solution — it won't cap the size of the content you're trying to send over.

thank you both 👍

PR #396, opened 2023-09-28
Implements bracketed paste in Kitty

gzagatti:
Kitty supports bracketed paste, but vim-slime did not have support for it. This PR adds bracketed paste to Kitty.

The implementation wraps the text with the bracketed paste escape characters `ESC [ 200 ~` an `ESC [ 201 ~` to obtain the desired result.

I've tested on my own setup and it works as expected.


jpalardy (approved):
looks familiar 👍

jpalardy:
@gzagatti thanks for the PR!

PR #401, opened 2023-11-03
kitty: get listen_on from environment if available

ptonner:
This checks for `KITTY_LISTEN_ON` as an environment variable and uses if set, which [ the docs](https://github.com/jpalardy/vim-slime/blob/main/assets/doc/targets/kitty.md) say is already done. But the code doesn't actually seem to read the environment variable. Also not sure why the docs suggests this will only work in vim, this works fine in nvim

This is also my first lines of vim script ever, so make sure I haven't done anything too stupid - namely I hope the check for empty string is all that's needed when setting `listen_on`

### DIFF
--- a/assets/doc/targets/kitty.md
+++ b/assets/doc/targets/kitty.md
@@ -34,7 +34,11 @@ listen_on unix:/tmp/mykitty
 
 ### SSH
 
-For slime use over ssh, you can also [forward the remote control](https://sw.kovidgoyal.net/kitty/kittens/ssh/#opt-kitten-ssh.forward_remote_control). Although note the additional security concerns: **this should only be done on trusted remote hosts**.
+For slime use over ssh, you can also [forward the remote control](https://sw.kovidgoyal.net/kitty/kittens/ssh/#opt-kitten-ssh.forward_remote_control).
+
+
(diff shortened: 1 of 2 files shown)


jpalardy:
Hi @ptonner 

Thanks for the PR :tada: 

I'm not a kitty user myself, nor have I written the original code. So, it's possible that either the code and/or the docs are in the best possible state.

It would seem to me that if line 4 was changed to:

```vim
let b:slime_config = {"window_id": 1, "listen_on": $KITTY_LISTEN_ON}
```

and nothing else was changed ... it would accomplish what's needed.

I don't know if you can weigh in, one way or another. We can iterate base on what you find.

I'm also reading this: https://sw.kovidgoyal.net/kitty/remote-control/#cmdoption-kitty-to -- so, it seems like `vim-slime` can delegate to kitty, without having to deal with `KITTY_LISTEN_ON` itself ... which was probably the original intent :thinking:

jpalardy:
It's looking good ^

If you're happy with this, I can merge this.

ptonner:
sure! I was just testing it out at work for the day, seems good to me

jpalardy:
Thanks again for the PR 👍

PR #403, opened 2023-11-30
Update kitty.vim

aloispichler:
provide additional hint for the desired variables


jpalardy:
Hi @aloispichler 

I'm not against this, but is it clear what the value in parentheses is?

Would provide a completion (similar [to this](https://github.com/jpalardy/vim-slime/blob/main/autoload/slime/targets/tmux.vim#L40-L44)) be a better fit for this?

aloispichler:
These fields are filled automatically, if slime is started in kitty and this terminal is used by slime as well. The text in brackets is intended to help the user to recall what to fill in here, if kitty is run in a different terminal.
Perhaps a text as `result of 'echo $KITTY_WINDOW_ID'` in the parenthesis is better.

jpalardy:
let's merge it, and we can iterate based on feedback?

Issue #408, opened 2024-01-13
slime_no_mappings works only if declaredbefore plugin is loaded


raffaem:
I'm putting `vim.g.slime_no_mappings = 1` to disable default mappings, as described in [advanced configuration guide](https://github.com/jpalardy/vim-slime/blob/main/assets/doc/advanced.md#advanced-configuration).

However, this only works if that line is declared BEFORE the plugin is loaded.

If declared after the plugin is loaded, it has no effect.

Should this be specified in the above mentioned guide?

jpalardy:
Hi @raffaem 

Generally, I think a lot of Vim plugins have a similar behavior.

Please send a documentation PR if you want to take a stab at it.

PR #410, opened 2024-01-14
Document slime_no_mappings

raffaem:


jpalardy:
Looking good, thanks ^

closes #408

PR #414, opened 2024-01-25
fix trailing whitespace in wezterm pane id

### LINKED ISSUE #413: WezTerm Pane Direction ID has trailing whitespace
Setting up slime with
```viml
 let g:slime_target = "wezterm"
 let g:slime_default_config = {"pane_direction": "right"}
```
will find the right pane id, but the pane id has trailing whitespace, so if one just presses enter on the suggested pane id, nothing will happen because WezTerm will not find that pane. Took me longer than I'd like to admit to figure out what the issue is. Removing the whitespace before accepting the suggested pane id works perfectly.

tbung:
Fixes #413.

### DIFF
--- a/autoload/slime/targets/wezterm.vim
+++ b/autoload/slime/targets/wezterm.vim
@@ -4,6 +4,7 @@ function! slime#targets#wezterm#config() abort
     let b:slime_config = {"pane_id": 1}
   elseif exists("b:slime_config.pane_direction")
     let pane_id = slime#common#system("wezterm cli get-pane-direction %s", [b:slime_config["pane_direction"]])
+    let pane_id = trim(pane_id)
     let b:slime_config = {"pane_id": pane_id}
   endif
   let b:slime_config["pane_id"] = input("wezterm pane_id: ", b:slime_config["pane_id"])


autoload/slime/targets/wezterm.vim
```diff
@@ -3,7 +3,7 @@ function! slime#targets#wezterm#config() abort
   if !exists("b:slime_config")
     let b:slime_config = {"pane_id": 1}
   elseif exists("b:slime_config.pane_direction")
-    let pane_id = slime#common#system("wezterm cli get-pane-direction %s", [b:slime_config["pane_direction"]])
+    let pane_id = trim(slime#common#system("wezterm cli get-pane-direction %s", [b:slime_config["pane_direction"]]))
```
jpalardy, line 6: looking good ^

can we put this on a separate line? (I think it'll be clearer)

```suggestion
    let pane_id = slime#common#system("wezterm cli get-pane-direction %s", [b:slime_config["pane_direction"]])
    let pane_id = trim(pane_id)
```

tbung: Done!

jpalardy:
Thanks for the PR 👍

Issue #420, opened 2024-03-18
neovim target TermOpen autocmd interactions with other terminal buffers


jmbuhr:
Would you be open to making the creation of this autocommand optional? As it also triggers on other terminal buffers that are not intended as a repl, like e.g. the terminal buffers created by neotest (https://github.com/nvim-neotest/neotest) to run tests or the terminal to run the preview command for quarto-nvim (https://github.com/quarto-dev/quarto-nvim/).

https://github.com/jpalardy/vim-slime/blob/51aaece8f7858358ca365466dec96c53b479e77a/plugin/slime.vim#L43-L49

In the case of neotest, this then leads to 

```
Error detected while processing TermOpen Autocommands for "*"..function slime#targets#neovim#SlimeAddChannel:
line    3:
E900: Invalid channel id
```

the first time a test is run.

`slime#targets#neovim#SlimeAddChannel(expand('<abuf>'))` could then be called manually on the creation of terminal intended for receiving code.

jpalardy:
@jam1015 ^

jam1015:
commented on PR #421, recommending merge. 

making this autocommand optional would require slightly in-depth changes.  For now I recommend both `g:slime_suggest_default = 0` and `g:slime_menu_config = 0`; that will remove the possibility of buffers that you don't want to send to's being suggested.  working on a fix to `SlimeClearChannel` analogous to the PR.

PR #421, opened 2024-03-18
fix: check that a terminal buffer is not hidden

jmbuhr:
before attempting to use it as a channel.
Otherwise `getbufvar` fails.

See https://github.com/jpalardy/vim-slime/issues/420

### DIFF
--- a/autoload/slime/targets/neovim.vim
+++ b/autoload/slime/targets/neovim.vim
@@ -111,6 +111,11 @@ endfunction
 
 function! slime#targets#neovim#SlimeAddChannel(buf_in) abort
   let buf_in = str2nr(a:buf_in)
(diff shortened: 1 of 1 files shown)


jpalardy:
@jam1015

jam1015:
Hi I just tested this like I did for my original PRs; 
with `nvim --clean -u ./minimal.lua` where minimal.lua was a minimal config file that loaded this plugin, and I fooled around with opening and closing terminals.  

This code doesn't break the plugin on that level of testing so I recommend merging it.

This has alerted me to other corner cases where non-standard buffers cause problems though, and I'm working out what those are and on fixes.

jpalardy:
Thanks for the PR 👍

PR #422, opened 2024-03-19
Fix "Terminal not found." when using together with `akinsho/toggleter…

you-n-g:
Fix "Terminal not found." when using together with `akinsho/toggleterm.nvim`

`akinsho/toggleterm.nvim` will not list its terminal in buffer list.
And it will return directly in this if
So it will results in that the toggleterm.nvim's terminal will not be added.


jpalardy:
@jam1015 ^

jam1015:
> @jam1015 ^

seen, will try to fix by tonight

jam1015:
fixed by pr #424

jpalardy:
@you-n-g can you confirm and close this PR?

you-n-g:
Thanks.

PR #426, opened 2024-03-30
When on Windows, check for cmd.exe

djhaskin987:
Neovim + slime gets tripped up on Windows when using conemu as the target. Neovim sets the shell slightly different, and the quotes don't work the same.

This PR fixes the problem.


autoload/slime/common.vim
```diff
@@ -35,7 +41,12 @@ function! slime#common#capitalize(text)
 endfunction
 
 function! slime#common#system(cmd_template, args, ...)
-  let escaped_args = map(copy(a:args), "shellescape(v:val)")
+  if &l:shell !=# "cmd.exe"
+      let escaped_args = map(copy(a:args), "shellescape(v:val)")
+  else
+      let escaped_args = a:args
+  endif
```
jpalardy, line 48: what's the story here?

is there never a need to escape for `cmd.exe`?

autoload/slime/common.vim
```diff
@@ -24,7 +24,13 @@ function! slime#common#write_paste_file(text)
...
     call mkdir(paste_dir, "p")
   endif
-  let output = slime#common#system("cat > %s", [slime#config#resolve("paste_file")], a:text)
+  if &l:shell ==# "cmd.exe"
+      let output = slime#common#system("type > %s", [slime#config#resolve("paste_file")], a:text)
+  elseif &l:shell ==# "powershell.exe"
+      let output = slime#common#system("echo > %s", [slime#config#resolve("paste_file")], a:text)
+  else
+      let output = slime#common#system("cat > %s", [slime#config#resolve("paste_file")], a:text)
+  endif
```
jpalardy, line 33: the `cat >` always felt like a hack…

I've been experimenting with [this](https://github.com/jpalardy/dotfiles/blob/main/vim/mappings.vim#L114). Can you give this a try and let me know if that works for you.

(untested) something like:

```vim
call writefile([a:text], slime#config#resolve("paste_file"))
```

jpalardy:
hi @djhaskin987 

Thanks for the PR, I added a few questions inline

autoload/slime/common.vim
djhaskin987: Rather, the escapes break cmd. It passes the literal quotes to the application and doesn't interpret them.

Sometimes. Maybe. :facepalm:

Anyway, this got it to work.

autoload/slime/common.vim
djhaskin987: I will give it a try, love it thanks!

djhaskin987: That line worked flawlessly! would recommend. I'll add it to my PR.

djhaskin987:
So, summary for this evening:
- `writefile` worked, I added it.
- In order to ensure I didn't mess up all the line endings in the repo while working on Windows I added a `.gitattributes` file.
- Left the if/then around CMD and escaped args unchanged.

jpalardy:
I'm not sold on the `.gitattributes` file, let's remove it for now.

The rest looks good 👍

djhaskin987:
OK, will do tonight :)

djhaskin987:
Sorry I didn't remove the attributes file I forgot.

jpalardy:
don't worry about it 👍 

I kept your commits in

Issue #427, opened 2024-04-23
Error raised after trigger send if there is no neovim terminal opened


jiz4oh:
<img width="748" alt="image" src="https://github.com/jpalardy/vim-slime/assets/41264693/3e5de124-09da-4753-9d68-4c16e4567a0a">

reproduce steps:
1. `nvim test.coffee`
2. call `:SlimeSendCurrentLine`

only occured when exists the `_EscapeText_{&filetype}` function

jpalardy:
cc @jam1015

jam1015:
I'm not able to reproduce this using a minimal config:

```lua
local lazypath = vim.fn.stdpath("data") .. "/lazy/lazy.nvim"
if not vim.loop.fs_stat(lazypath) then
  vim.fn.system({
    "git",
    "clone",
    "--filter=blob:none",
    "https://github.com/folke/lazy.nvim.git",
    "--branch=stable", -- latest stable release
    lazypath,
  })
end
vim.opt.rtp:prepend(lazypath)


require("lazy").setup(
  {
    {
      "jam1015/vim-slime",
      init = function()
        require("plugin_configs.vim-slime.initi")


        vim.g.slime_target = "neovim"
        vim.g.slime_no_mappings = true
        vim.g.slime_input_pid = false
        vim.g.slime_suggest_default = true

        vim.g.slime_menu_config = false
      end,

      config = function()
        local keymap = vim.keymap.set
        keymap("n", "gz", "<Plug>SlimeMotionSend", { remap = true, silent = false })
        keymap("n", "gzz", "<Plug>SlimeLineSend", { remap = true, silent = false })
        keymap("x", "gz", "<Plug>SlimeRegionSend", { remap = true, silent = false })
      end,
    },
  }

)
```

I run `nvim --clean -u ./minimal.lua  test.coffee` and then `:SlimeSendCurrentLine` with no terminal open and just get the `Terminal not found.` warning message (I get it twice, this is a bug that will take some refactoring to fix, but that isn't relevant here).   The text is successfully sent when I open a terminal.  

Also tested on an example python file which also has an `_EscapeText_python` function and had no problems.

Maybe try adding your personal config piece by piece to the minimal config to see what causes the errors.

jiz4oh:
[...]

PR #428, opened 2024-04-24
addresses issue #427

jam1015:
Issue #427 shows that ugly error messasges appear for certain filetype plugins  if they break the text into multiple chunks, and the config is already not valid, for the neovim target.  This was caused by the Neovim send function unletting invalid configs.  To fix it, it now just sets the config to an empty dictionary.


jpalardy:
thanks for the quick fix 👍

ppalmes:
please double check that this change does not affect vim because i experiences many issues with extra characters in the newline in shell files and concatenating multiple lines as one.

jam1015:
@ppalmes this won't affect vanilla vim, only neovim, my guess is that you should raise your own issue with a minimal example that causes the problem

PR #433, opened 2024-05-14
Rework screen target

jpalardy:
Removed dependency on `g:slime_paste_file` by using STDIN with screen `readreg`.

Advantages

- less steps
- no touching the filesystem (faster, probably)
- no leaving .slime_paste file around

It might break something, in which case I'll see if when need to support both STDIN and `g:slime_paste_file`

related: #431

### DIFF
--- a/assets/doc/targets/screen.md
+++ b/assets/doc/targets/screen.md
@@ -7,16 +7,6 @@ By default, [GNU Screen](https://www.gnu.org/software/screen/) is assumed, you d
 let g:slime_target = "screen"
 ```
 
-Because Screen doesn't accept input from STDIN, a file is used to pipe data
-between Vim and Screen. By default this file is set to `$HOME/.slime_paste`.
-The name of the file used can be configured through a variable:
-
-    let g:slime_paste_file = expand("$HOME/.slime_paste")
-    " or maybe...
-    let g:slime_paste_file = tempname()
-
-⚠️  This file is not erased by the plugin and will always contain the last thing you sent over.
-
 When you invoke `vim-slime` for the first time, you will be prompted for more configuration.
 
 screen session name:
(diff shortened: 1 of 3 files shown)


jpalardy:
Hi @jam1015 

This `paste_file` stuff had bothered me for a while.

It seems the last use is [in the neovim target](https://github.com/jpalardy/vim-slime/blob/main/autoload/slime/targets/neovim.vim#L104) … and might not even be needed? 🤔   
(or we use it for the wrong reason…)

Let me know if you think we can remove it.

jam1015:
Hi!  After investigating and testing with a minimal config, I think you're good to get rid of this line. It doesn't seem to be accessed anywhere. 

The comment for it:

```
    " Neovim jobsend is fully asynchronous, it causes some problems with
    " iPython %cpaste (input buffering: not all lines sent over)
    " So this `write_paste_file` can help as a small lock & delay
    " call slime#common#write_paste_file(a:text)

```
indicates that it is used to introduce a small delay and that delay in introduced by  [the python ftplugin file](https://github.com/jpalardy/vim-slime/blob/main/ftplugin/python/slime.vim) when `slime_python_ipython` is set.  So  maybe using the pastefile for  this is a relic from  a previous strategy used to introduce a delay.

jpalardy:
Do you want me to take a stab at it?

I can code it, but I don't know how thorough my neovim tests will be 😅

jam1015:
Ok go ahead and make the changes you think are appropriate.  The changes to be made in my opinion are just, deleting the call to ` call slime#common#write_paste_file(a:text)` and the comment explaining that call, and if its then not used anywhere else in the codebase, deleting the definition of that function. Once you've done that I can test that branch; my tests are just sending code to python/ipython/ptpython in neovim with `g:slime_python_ipython` enabled and disabled.

PR #435, opened 2024-05-17
Remove paste_file from neovim target

jpalardy:
Removed dependency on `g:slime_paste_file` which didn't even look like it was needed 🤔 

Advantages

less steps
- no touching the filesystem (faster, probably)
- no leaving `.slime_paste` file around

related: https://github.com/jpalardy/vim-slime/pull/431

### DIFF
--- a/autoload/slime/targets/neovim.vim
+++ b/autoload/slime/targets/neovim.vim
@@ -98,10 +98,6 @@ function! slime#targets#neovim#send(config, text) abort
   " existence of 'b:config' is implicitly assumed by the base send function
   " if it doesn't exist there will have been an error before reaching this code
   if slime#targets#neovim#ValidConfig(a:config,0)
-    " Neovim jobsend is fully asynchronous, it causes some problems with
-    " iPython %cpaste (input buffering: not all lines sent over)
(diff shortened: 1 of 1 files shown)


jpalardy:
@jam1015 

I very lightly tested this, as neither a neovim nor python user.

I got burned on #434 because I had misconfigured and wasn't even testing my changes 😬 

As a safety net, I used `chmod 000 ~/.slime_paste` so that any attempt to read or write will blow up.

jam1015:
@jpalardy I just tested it and it seems to work on the neovim both for ipython (mentioned in the comment justifying the paste file) and for the vanilla python REPL, and for R. tried indented code blocks for each.    ipython needs `g:slime_python_ipython` and bracketed paste doesn't work but I think that is part of the neovim target.   ptpython, another python REPL that I like, doesn't work in the neovim target and seems to need bracketed paste.  I don't know if a pastefile can help with that, but, since we've established that it isn't even used by the Neovim target, I would merge this PR.

jpalardy:
thanks for confirming 🙏 

we'll see what happens  😄

PR #424, opened 2024-03-20
Better dealing with hidden and non-interactive terminals

jam1015:
Over the past two days  issue #420, pr #421, and  pr #422 dealt with how the Neovim target handles non-interactive terminals, and unlisted terminals.

But pr #421 and #422 are in conflict with each other. I aimed to fix the problems that those PRs addressed and resolve the conflict between them by creating an additional user-configurable option.

This PR:

- adds a check in `SlimeAddChannel` to make sure that a terminal that is opened has the `&channel` setting, which is what determines if it is a valid target.  I did a quick test with [neotest](https://github.com/nvim-neotest/neotest) and as mentioned in issue #420 , following the example tests in [this tutorial](https://semaphoreci.com/community/tutorials/testing-python-applications-with-pytest), and an error as described came up when this change wasn't added, and was fixed when it was.

- adds a configurable option , `g:slime_neovim_ignore_unlisted` that turns on or off unlisted buffers being available as targets. This addresses PR #422 

- adds documentation of these changes

PR #419, opened 2024-02-27
Validate in target

jam1015:
In my [recent issue](https://github.com/jpalardy/vim-slime/issues/417) I proposed changes to `autoload/slime.vim`.  You suggested a more conservative approach.  The two approaches are summarized in this image:
image
With your preferred way on the left, and what I suggested on the right, with your annotations overlaid.

I said I'd open two pull requests, one with your preferred way and one with my preferred way.  Instead I'm just opening this one, which is your preferred way; the changes to the Neovim target code are quite extensive (in a way that I feel is justified) so it will be easier to try to get this merged first, because the changes here are already quite extensive.

I know that this makes the Neovim target the one with the most code by far, but I think what I have here enhances the user experience overall. Give it a try/compare it to how it was before to sew why I made it this way.

Here is an explanation of the changes:

## `README.md`

A small change to phrasing.

## `plugin/slime.vim`

`slime#targets#neovim#SlimeAddChannel(expand('<abuf>'))` and `slime#targets#neovim#SlimeClearChannel(expand('<abuf>'))` use `<abuf>` now to track exactly the buffers being acted on regardless of whether they are active.  I didn't know about this feature before and it allowed me to simplify these channel tracking functions/not have to deal with corner cases.

## `autoload/slime.vim`

- Made a small change here, adding a a check that the config is not empty.  In this case it helpe to make sure the user is prompted for a config when there is an attempt to send text the current config is empty.  This is a very light version of my ultimate desired changes, which will catch more cases. 

- Removed redundant calls to `GetConfig`. With them in, we are sometimes prompted to configure twice if we mmisconfigure the first time.

## `assets/doc/targets/neovim.md`

A full explanation of the features I added to `autoload/slime.vim` (read it to see everything I have added).  This is a lot but I commit to maintaining it.

The goal here is to give very clear documentation so that someone with late-beginner vim knowledge can get started more easily.

I found that these changes were difficult to implement using overrides, hence this PR.

## `autoload/slime/targets/neovim.vim`

Here are the big changes.

Calls to my validation functions are added by way of `s:protected_validation_and_clear` which checks for existence of a buffer's config, and then makes sure that it is correct. If the config is not valid, all identical configs in other buffers are cleared.

### `slime#targets#neovim#config()`

Wrapped in a call that checks for a valid environment.

Changed to allow config with a menu prompt or direct entry on the command line (configurable as detailed in the Neovim documentation).

Still allows users to enter `pid` or `jobid`.

[...]

PR #436, opened 2024-05-22
Main to merge

jam1015:
In PR #419 
I end my explanation with:

> If you accept this, my further desired changes to autoload/slime.vim will allow this code to be simplified quite a bit.
> 
> Right now are also some double warning messages, and corner cases where you are prompted to config twice if you misconfigure the first time, and these are also fixed by my modifications to `autoload/slime.vim` and other files that I have in another branch.

This PR has the desired changes that I talked about.

Currently functions in the neovim target deal with what they are supposed to do, and also with validation.  This PR splits up those functions,  making dedicated validation functions in  the neovim target and putting the logic of how to call them in `autoload/slime.vim`.  These changes in `autoload/slime.vim` are compatible with targets that don't have the validation functions I added to Neovim, and the code for those targets could be expanded in the future to include validation functions.   I've also confirmed that code from PR #424 is preserved, and that the problems identified in issue #427 don't resurface.

Besides cleaning the code, the improvement is that if someone misconfigures the neovim target, an accurate warning message of just one line is displayed. previously only the 'terminal not found' warning message was displayed, multiple times for the same error.   This is because the 'try catch' blocks in `autoload/slime.vim` stops multiple errors from firing.


jpalardy:
I'll have time to review this tonight

jpalardy:
Hi @jam1015 

While I'm happy to delegate the code for `neovim`, I'm concerned about any changes to the "core".

It seems like this is backward compatible.

Is there no way forward without touching the core? (if not, I'll need to spend more time thinking about this)

jam1015:
Hi! Thanks so much for taking the time to review it.  

I tried for a long time to not touch the core but to make it work like it in my opinion should I had to but wasn't able to.  


Consider `slime#send` before my changes:

```
function! slime#send(text)
  call s:SlimeGetConfig()

  " this used to return a string, but some receivers (coffee-script)
  " will flush the rest of the buffer given a special sequence (ctrl-v)
  " so we, possibly, send many strings -- but probably just one
  let pieces = s:_EscapeText(a:text)
  for piece in pieces
    if type(piece) == 0  " a number
      if piece > 0  " sleep accepts only positive count
        execute 'sleep' piece . 'm'
      endif
    else
      call s:SlimeDispatch('send', b:slime_config, piece)
    endif
  endfor
endfunction
```



If something goes wrong in `s:SlimeGetConfig` and is caught in the functions in the targets, it still moves ahead and tries to send.   In my Neovim code as it is in your version or the repo now, a nice error message is printed if the user misconfigures, `slime#send()` but it tries to dispatch the target send function anyway, and a second error message is printed, in the  case of Neovim indicating an empty config;  the code sets the config to an empty dictionary when there is an error in configuration.   This is confusing to the user; they just put in a configuration, why is it empty all of the sudden?   Just one error message describing what went wrong is better.

with my changes:



```

function! slime#send(text)
  if s:SlimeDispatchValidate("ValidEnv")
    try
      call s:SlimeGetConfig()
    catch \invalid config\
      return
    endtry

    " this used to return a string, but some receivers (coffee-script)
    " will flush the rest of the buffer given a special sequence (ctrl-v)
    " so we, possibly, send many strings -- but probably just one
    let pieces = s:_EscapeText(a:text)
    for piece in pieces
      if type(piece) == 0  " a number
        if piece > 0  " sleep accepts only positive count
          execute 'sleep' piece . 'm'
        endif
      else
        call s:SlimeDispatch('send', b:slime_config, piece)
      endif
    endfor
  endif
endfunction
```

If the user misconfigures, the target configuration function prints an explanatory warning, throws 'invalid config' and it is caught in the core function, and the core function returns, and nothing gets sent. A single error message, whatever is printed by the code that did the throwing is printed and the user sees exactly what went wrong. 
 


I can probably test more and think of more examples but they are all similar to this.



There is also the general benefit of cleaner code in the neovim target; each function has one purpose, not a purpose + validation.  

These benefits could be extended to other targets; yes I know they all have much less code than the neovim target, but a target could come along one day for which these changes are helpful. 


- A criticism that I do have of my code is that it uses a try catch block to catch `error 117 function does not exist`, to ensure backward compatibility.  I know that relying on try-catch (or exceptions or something of the like) frowned upon in some programming languages like c++ but that is the best I can do here. See the comment in the body of `function! s:SlimeDispatchValidate(name, ...)
`  for a justification.

jpalardy:
Hi @jam1015 

Let me do another pass at this when I have more time.

autoload/slime.vim
```diff
@@ -122,30 +138,61 @@ endfunction
...
+  " using try catch because exists() doesn't detect autoload functions that aren't yet loaded
+  " the idea is to return the interger 1 for true in cases where a target doesn't have
+  " the called validation function implemented. E117 is 'Unknown function'.
+  try
+    return call(fun_string, a:000)
+  catch /^Vim\%((\a\+)\)\=:E117:/
+    return 1
+  endtry
+
+endfunction
```
jpalardy, line 194: I'm warming up to this ^

A few ideas:
- what if we remove line 178-184, and delegate to `s:SlimeDispatch(name, ...)` below. It seems most of this logic is duplicated, no?
- we keep the try-catch wrap
- maybe rename to `s:SlimeDispatchSafe`? (I'm open to a better name)

```vim
# caller
if exists("b:slime_config") && !s:SlimeDispatchSafe("ValidateConfig", b:slime_config, 0)
  ...
```

```vim
# (untested) idea
function! s:SlimeDispatchSafe(name, ...)
  " using try catch because exists() doesn't detect autoload functions that aren't yet loaded
  " the idea is to return the interger 1 for true in cases where a target doesn't have
  " the called validation function implemented. E117 is 'Unknown function'.
  try
    return s:SlimeDispatch(name, a:000)
  catch /^Vim\%((\a\+)\)\=:E117:/
    return 1
  endtry
endfunction
```

jam1015: Hi! Sorry for disappearing for a few months.   I essentially took your suggestion and now have:

```vim
function! s:SlimeDispatchValidate(name, ...)
  " using try catch because exists() doesn't detect autoload functions that aren't yet loaded
  " the idea is to return the interger 1 for true in cases where a target doesn't have
  " the called validation function implemented. E117 is 'Unknown function'.
  try
    return call ("s:SlimeDispatch", [a:name] + a:000)
  catch /^Vim\%((\a\+)\)\=:E117:/
    return 1
  endtry
endfunction

function! s:SlimeDispatch(name, ...)
  " allow custom override
  let override_fn = "SlimeOverride" . slime#common#capitalize(a:name)
  if exists("*" . override_fn)
    return call(override_fn, a:000)
  endif
  return call("slime#targets#" . slime#config#resolve("target") . "#" . a:name, a:000)
endfunction
```

I still call it `s:SlimeDispatchValidate` because its purpose is still validation (hence the default return value of 1 for functions that don't exist). 

I've manually tested this in on the neovim target, the tmux target from both vim and neovim, and the vimterminal target.

autoload/slime/targets/neovim.vim
```diff
@@ -348,14 +307,13 @@ function! s:config_with_menu() abort
...
-  let b:slime_config = {}
+    let b:slime_config = {}
     return
   endif
 
   let used_config = term_bufinfo[selection - 1]
 
-  let config_out = {"jobid": used_config["jobid"], "pid": used_config["pid"] }
-  return config_out
+  return {"jobid": used_config["jobid"], "pid": used_config["pid"] }
```
jam1015, line 316: removed unnecessary  line

PR #444, opened 2024-09-01
Added a paste guard for tmux

### LINKED ISSUE #443: tmux empty paste corner case
When sending a blank line to tmux (e.g. `slime#send('')`), it instead sends the contents of the last tmux yank register.

This is because [the buffered for loop](https://github.com/jpalardy/vim-slime/blob/f6bbbeb8f18393f2177cdcef764f3e0e1d7e9328/autoload/slime/targets/tmux.vim#L23-L31) has a `range(0,0)` (i.e `[0]`) command in this case, which executes one loop, which loads nothing into the buffer (which does nothing), and then sends the last non-empty thing that was loaded.

In my use case this has involved sending some relatively dangerous commands to a terminal!

This can be fixed by wrapping the loop in a `if len(text_to_paste) >0` statement, but wanted to check if

a) Is this a feature or a bug?
b) Is there a neater/more robust way to achieve this?

williamjameshandley:
Resolves #443

### DIFF
--- a/autoload/slime/targets/tmux.vim
+++ b/autoload/slime/targets/tmux.vim
@@ -20,15 +20,17 @@ function! slime#targets#tmux#send(config, text)
   " reasonable hardcode, will become config if needed
   let chunk_size = 1000
 
-  for i in range(0, len(text_to_paste) / chunk_size)
-    let chunk = text_to_paste[i * chunk_size : (i + 1) * chunk_size - 1]
-    call slime#common#system(target_cmd . " load-buffer -", [], chunk)
-    if bracketed_paste
-      call slime#common#system(target_cmd . " paste-buffer -d -p -t %s", [a:config["target_pane"]])
-    else
-      call slime#common#system(target_cmd . " paste-buffer -d -t %s", [a:config["target_pane"]])
-    end
-  endfor
+  if len(text_to_paste) > 0
+    for i in range(0, len(text_to_paste) / chunk_size)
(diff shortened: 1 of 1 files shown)


autoload/slime/targets/tmux.vim
```diff
@@ -20,15 +20,17 @@ function! slime#targets#tmux#send(config, text)
...
-  for i in range(0, len(text_to_paste) / chunk_size)
-    let chunk = text_to_paste[i * chunk_size : (i + 1) * chunk_size - 1]
-    call slime#common#system(target_cmd . " load-buffer -", [], chunk)
-    if bracketed_paste
-      call slime#common#system(target_cmd . " paste-buffer -d -p -t %s", [a:config["target_pane"]])
-    else
-      call slime#common#system(target_cmd . " paste-buffer -d -t %s", [a:config["target_pane"]])
-    end
-  endfor
+  if len(text_to_paste) > 0
```
jpalardy, line 23: I fixed this by adding

```vim
  if len(text_to_paste) == 0
    return
  end
```

after line 15 above

can you make that change?

(also, see my comment in #443, I'm having reproducing this)

jpalardy:
No response; addressed in [#c8cfb9bec](https://github.com/jpalardy/vim-slime/commit/c8cfb9bec587aef0a7a87a52dacb087f48ea2b41)

Please re-open if it's still a problem

closes #443

PR #446, opened 2024-09-11
Use the paste file in conemu for faster turnaround

djha-skin:
This PR changes conemu to use the paste file instead of the windows clipboard. This can make it slightly faster if the clipboard is configured to use a subprocess like powershell.

Changes:

* The `slime#common#write_paste_file` now returns the name of the paste file to which it wrote, so that `slime#resolve#config` doesn't have to be called twice, once in the `conemu#send` and once in the `write_paste_file` method.
* The case where the paste file isn't written is checked.
* The paste file has the `S` flag set, which tells vim not to bother calling `fsync()` on the file, for faster write speeds. This makes sense. The file doesn't contain anything that needs to persist across reboots.
* ConEMU writes to the paste file and uses that instead.

This change has been "tested on my machine" and works.

### DIFF
--- a/autoload/slime/common.vim
+++ b/autoload/slime/common.vim
@@ -25,7 +25,12 @@ function! slime#common#write_paste_file(text)
     call mkdir(paste_dir, "p")
   endif
   let lines = split(a:text, "\n", 1)
-  call writefile(lines, slime#config#resolve("paste_file"), 'b')
+  let paste_file = slime#config#resolve("paste_file")
+  let result = writefile(lines, paste_file, 'bS')
+  if result != 0
+    echoerr "Couldn't write to slime paste file."
+  endif
+  return paste_file
 endfunction
 
 function! slime#common#capitalize(text)
--- a/autoload/slime/targets/conemu.vim
+++ b/autoload/slime/targets/conemu.vim
@@ -9,11 +9,8 @@ function! slime#targets#conemu#config() abort
 endfunction
 
 function! slime#targets#conemu#send(config, text)
-  " use the selection register to send text to ConEmu using the windows clipboard (see help gui-clipboard)
-  " save the current selection to restore it after send
-  let tmp = @*
-  let @* = a:text
-  call slime#common#system("conemuc -guimacro:%s print", [a:config["HWND"]])
-  let @* = tmp
+  " Use the selection register to send text to ConEmu using the slime paste file
+  let paste_file = slime#common#write_paste_file(a:text)
+  call slime#common#system("conemuc -guimacro:%s pastefile 2 %s", [a:config["HWND"], paste_file])
 endfunction
 


autoload/slime/common.vim
```diff
@@ -25,7 +25,12 @@ function! slime#common#write_paste_file(text)
     call mkdir(paste_dir, "p")
   endif
   let lines = split(a:text, "\n", 1)
-  call writefile(lines, slime#config#resolve("paste_file"), 'b')
+  let paste_file = slime#config#resolve("paste_file")
+  let result = writefile(lines, paste_file, 'bS')
+  if result != 0
+    echoerr "Couldn't write to slime paste file."
+  endif
+  return paste_file
```
jpalardy, line 33: my initial reaction was `resolve` is probably too fast to benchmark (it's not in some inner loop) 🤔 

but looking at the call sites, it wouldn't hurt to return `paste_file` -- so I think I'm OK with that 👍 

I like checking for a `writefile` error, how about avoid the `result` var? (it's only used for the `if`)

```suggestion
  let paste_file = slime#config#resolve("paste_file")
  if writefile(lines, paste_file, 'bS') != 0
    echoerr "Couldn't write to slime paste file."
  endif
  return paste_file
```

autoload/slime/targets/conemu.vim
```diff
@@ -9,11 +9,8 @@ function! slime#targets#conemu#config() abort
...
 function! slime#targets#conemu#send(config, text)
-  " use the selection register to send text to ConEmu using the windows clipboard (see help gui-clipboard)
-  " save the current selection to restore it after send
-  let tmp = @*
-  let @* = a:text
-  call slime#common#system("conemuc -guimacro:%s print", [a:config["HWND"]])
-  let @* = tmp
+  " Use the selection register to send text to ConEmu using the slime paste file
+  let paste_file = slime#common#write_paste_file(a:text)
+  call slime#common#system("conemuc -guimacro:%s pastefile 2 %s", [a:config["HWND"], paste_file])
```
jpalardy, line 14: all in all, I'm surprised that using the pastefile would be _faster_ than the clipboard … 🤔 

---

> The paste file has the S flag set, which tells vim not to bother calling fsync() on the file
…
> The file doesn't contain anything that needs to persist across reboots.

That wasn't my understanding of `fsync` — doesn't that create a race-condition if we
1. write (but skip flush)
2. turn around and read the file

?

---

another concern is that the pastefile sticks around, holding on to your last slime send — I'm not sure everybody feels comfortable with that, so I've been avoiding it when possible (only `screen` was still using it, because there were no other choice 😮‍💨 )

jpalardy (commented):
hI @djhaskin987 

thanks for the PR — I've left some comments inline

jpalardy:
@djhaskin987 

any thoughts on this?

PR #458, opened 2025-04-20
Documentation Fixes/Bugfix

jam1015:
This pull request does two things

1) Fixes small typos/slightly expands neovim target documentation

2) Fixes a bug in the automatic configuration with `vim.g.slime_get_jobid`; was getting a type error I hadn't caught before with using a function type in an if-statement.


jpalardy:
Hi @jam1015 

Looking good, thanks 👍

PR #461, opened 2025-04-21
Shallow-copy default config to buffer variable

thejohnfreeman:
Because assignment in Vimscript copies references, the old behavior causes two buffers to share the same config dictionary object if one sets `g:slime_default_config`. This new behavior guarantees that they create unique copies of the default.

### DIFF
--- a/autoload/slime.vim
+++ b/autoload/slime.vim
@@ -35,7 +35,7 @@ function! s:SlimeGetConfig()
   " assume defaults, if they exist
 
   if exists("g:slime_default_config")
-    let b:slime_config = g:slime_default_config
+    let b:slime_config = copy(g:slime_default_config)
     if exists("b:slime_config") && !s:SlimeDispatchValidate("ValidConfig", b:slime_config, 0)
         unlet b:slime_config
     endif


jpalardy:
Hi @thejohnfreeman 

Good catch 👍 

It's true that I don't usually modify `g:slime_default_config` outside of setting it in my `.vimrc`; it's a "constant" to me.

I didn't check the code, but are there other places that could use the same `copy(…)` fix?

thejohnfreeman:
Oh, I see how my wording was ambiguous. I meant "one" as in "oneself", not "one of the two buffers". I set it as a constant in my `.vimrc` too. But this makes it difficult to configure two buffers to send text to different panes in tmux. The current plugin behavior assigns a reference to `g:slime_default_config` as `b:slime_config`, so the two buffers share the same config and the last modification wins.

thejohnfreeman:
I haven't looked for other troublesome assignments. I only use tmux, and this was enough to fix my personal issue. It's possible the other targets have the same problem.

jpalardy:
Alright, I'm all for merging this.

Let me do a quick pass to see if other places also need the fix.

jpalardy:
I found other places, but I don't think they are relevant.

PR #466, opened 2025-08-21
Add bracketed-paste support for neovim

frankier:
Neovim's terminal supports this, so I've added it following the other implementations.

### DIFF
--- a/assets/doc/targets/neovim.md
+++ b/assets/doc/targets/neovim.md
@@ -193,6 +193,19 @@ local function get_slime_pid()
 end
 ```
 
+### bracketed-paste
+
+Some REPLs can interfere with your text pasting. The [bracketed-paste](https://cirw.in/blog/bracketed-paste) mode exists to allow raw pasting.
+
+`neovim` supports bracketed-paste, use:
+
+```vim
+let g:slime_bracketed_paste = 1
+" or
+let b:slime_bracketed_paste = 1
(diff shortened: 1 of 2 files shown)


jpalardy:
Hi @frankier 

Thanks for the PR 👍 

I'm tried it and I see this (using R)

Screenshot 2025-08-21 at 10 06 12

Can you tell me if you can reproduce this? How are you configured?

frankier:
It works fine for me in a few settings with NVIM v0.10.4. It works pasting into R, Julia and bash. I also tried running within xterm instead of Alacritty. How is your R configured?

jpalardy:
I'm using Ghostty, NVIM 0.11.3, R 4.5.1

if I'm trying to simplify this to the smallest bit, `call chansend(3, "\e[200~")` (where `3` is the jobid) also reproduces the problem

relevant nvim config

```
vim.g.slime_target = "neovim"
vim.g.slime_bracketed_paste = 1
```

…

interesting … I see it happening in R and irb … but not python 🤔 

so, it must be something about how it got compiled/installed

can you do a few quick checks
- does `call chansend` (above) work for you?
- can you reproduce this in some REPL?

jpalardy:
Yes, that sounds about right.

Let's merge this and, depending on what issues come up, I might have to loop you in.

Thanks again!

PR #468, opened 2025-08-23
Support lazyloading by existing terminals for neovim

frankier:
If this plugin is lazy loaded in response to a keybinding e.g. using lazy.nvim then there can be existing terminals it doesn't know about. This PR just adds the terminals from all existing buffers at startup.

P.S. Thanks a lot for the plugin and being so responsive. Very much appreciated!

### DIFF
--- a/plugin/slime.vim
+++ b/plugin/slime.vim
@@ -40,6 +40,10 @@ endif
 " the channel id of the last opened terminal
 if slime#config#resolve("target") == "neovim"
   if has('nvim')
+    " Add existing buffers (SlimeAddChannel will check if it's a terminal)
+    for buf in getbufinfo()
+      call slime#targets#neovim#SlimeAddChannel(buf.bufnr)
+    endfor
     augroup nvim_slime
       autocmd!
       " keeping track of channels that are open


jpalardy:
Hi @frankier 

I'm not too in love with adding move neovim-specific logic in `plugins/slime.vim`

is there another way to do this? 🤔

frankier:
In principle `vim.g.slime_get_jobid` acts as an "escape hatch" which would allow terminals to be selected in another way e.g. if snacks.nvim was being used then https://github.com/folke/snacks.nvim/blob/main/docs/terminal.md#snacksterminallist could be used to find the terminals. Unfortunately, the current logic tries to keep track of what terminals are available and guards against the case of no terminals, here:

https://github.com/jpalardy/vim-slime/blob/48704b357d9ec7b333c4ed856b71dc6ecf24d48d/autoload/slime/targets/neovim.vim#L123-L129

The problem is that even if people overload `vim.g.slime_get_jobid` the current "keeping track of loaded terminals" logic and accompanying guard still runs. I don't particularly want the current behaviour and would prefer to just override `vim.g.slime_get_jobid`, but the PR as it stands allows execution to reach `vim.g.slime_get_jobid` and should avoid breaking the current behaviour.

One possibility is we could decide to always return `true` from `slime#targets#neovim#ValidEnv` if `vim.g.slime_get_jobid` is defined. Perhaps keeping track of the terminals could also be disabled in this case? Would there be some unintended consequences of this or would it be cleaner?

Maybe @jam1015 could chime in here, since I understand that they've been the main one working on the neovim target so far.

jam1015:
Ok I implemented this in #469

